PR #168, opened 2023-08-29
fix(shipments): prevent error in barcode note identifier

joerivanveen:


config/pdk-settings.php
```diff
@@ -116,7 +116,7 @@
      */
 
     'createBarcodeNoteIdentifier' => factory(function () {
-        return static function (int $shipmentId) {
+        return static function (string $shipmentId) {
```
EdieLemoine, line 119: ik mis tests

EdieLemoine (changes requested):

EdieLemoine (approved):

PR #319, opened 2024-11-13
feat: allow receipt code to be used for country BE

GravendeelJochem:
INT-713


FlorianSDV (commented):
Dit ziet er prima uit.

PR #355, opened 2025-05-20
fix(settings): remove save address option when unavailable

joerivanveen:
Saving the client address in the backoffice address book is not possible in order modus. The toggle being still present confused users obviously. Removed the toggle when the option is not available.

Reference: https://developer.myparcel.nl/api-reference/14.order-object-definitions.html

INT-902


joerivanveen:
> ## Pull Request Overview
> This PR removes the "save customer address" toggle from the Order Settings view when the order mode does not support saving addresses.
> 
> * Removed the SAVE_CUSTOMER_ADDRESS toggle when the feature is unavailable, using a visibleWhen condition.

That’s what I said :-)

GravendeelJochem (approved):

FreekVR (approved):

PR #371, opened 2025-07-18
fix: replace UPS carrier 8 with carriers 12 and 13

GravendeelJochem:
INT-903


src/App/Order/Calculator/UPSStandard/UPSStandardDeliveryTypeCalculator.php
```diff
@@ -0,0 +1,25 @@
...
+
+declare(strict_types=1);
+
+namespace MyParcelNL\Pdk\App\Order\Calculator\UPSStandard;
+
+use MyParcelNL\Pdk\App\Order\Calculator\AbstractPdkOrderOptionCalculator;
+use MyParcelNL\Pdk\Shipment\Model\DeliveryOptions;
+use MyParcelNL\Pdk\Types\Service\TriStateService;
+
+final class UPSStandardDeliveryTypeCalculator extends AbstractPdkOrderOptionCalculator
```
FreekVR, line 11: Refer to my comment for UPS express, it also applies here

src/App/Order/Calculator/UPSStandard/UPSStandardInsuranceCalculator.php
```diff
@@ -0,0 +1,23 @@
...
+    public function calculate(): void
+    {
+        $shipmentOptions = $this->order->deliveryOptions->shipmentOptions;
+
+        // Set default insurance to 10000 (100 euro) if not set
+        if (!isset($shipmentOptions->insurance) || $shipmentOptions->insurance === 0) {
+            $shipmentOptions->insurance = 10000;
+        }
+    }
+}
```
FreekVR, line 23: Missing newline

FreekVR (changes requested):
I think this would introduce a number of breaking changes due to there being no mapping from the "old" to "new" UPS and removing all the existing UPS config.

This potentially breaks:
- Existing settings applied to the old UPS would not automatically carry over to the new UPS definitions
- Existing orders placed with the UPS carrier selected would likely not be exportable
- Orders which were already exported before may not show the correct information or be editable through the admin

FreekVR (approved):

PR #377, opened 2025-08-04
feat(proposition): proof of concept for implementing a shared proposition config

FreekVR:
adds a new concept for a proposition config, to replace existing platform functionality in the future

related to INT-1084


src/Proposition/Service/PropositionService.php
```diff
@@ -0,0 +1,332 @@
...
+                'outboundFeatures' => $contract['outboundFeatures'] ?? [],
+                'inboundFeatures' => $contract['inboundFeatures'] ?? [],
+            ];
+            $carrierModels[] = new Carrier($carrierData);
+        }
+
+        // Combine with carrier-specific own contracts
+        foreach ($this->getPropositionConfig()->contracts->availableForCustomCredentials as $customContract) {
+            // Skip already-defined carriers
+            if (in_array($customContract['carrier']['id'], array_column($carrierModels, 'id'))) {
```
NabDevs, line 142: Just checking, will `array_column($carrierModels, 'id')` work as expected here? Since `Carrier` stores values in `$attributes`(protected), I’m not sure if `array_column()` will pick it up or am I missing something?

NabDevs (approved):
Awesome job!

joerivanveen (changes requested):

FreekVR (commented):
S

tests/Unit/Base/AbstractPdkBootstrapperTest.php
```diff
@@ -93,16 +95,19 @@ protected function getAdditionalConfig(
...
     factory(Account::class, $platformId)
         ->withShops()
         ->store();
 
-    expect(PlatformFacade::getPropositionName())->toBe($platform);
+    expect($propositionService->getActivePropositionId())
+        ->toBe($platformId)
+        ->and($propositionService->getPropositionConfig()->proposition->key)
+        ->toBe($platform);
 })->with([
```
joerivanveen, line 110: Why did you remove this test: `'unrecognized id' => [-100, Platform::MYPARCEL_NAME],`, it worries me that peoples plugin / pdk can become unusable when, for whatever reason, there is a faulty platformId in their account? I understand that you want to change this behaviour in the future but for now an error would prevent people from filling in their api key, if I’m not mistaken.

FreekVR: Yes and that is exactly the goal, as we do not want to default a customer from eg. France (Platform ID 7 lets say) to a proposition (ID 1) that is not correct for them, perhaps without them every realizing they are sending with Dutch carriers (and getting into trouble much further down the line)

This behavior has coverage in the PropositionServiceTest

joerivanveen (approved):

PR #382, opened 2025-08-14
feat: add acceptance api

GravendeelJochem:
INT-891

### DIFF
--- a/src/App/Action/Backend/Account/UpdateAccountAction.php
+++ b/src/App/Action/Backend/Account/UpdateAccountAction.php
@@ -104,17 +104,11 @@ class UpdateAccountAction implements ActionInterface
     {
         // Get existing account settings to preserve them
         $existingSettings = $this->pdkSettingsRepository->all()->account;
-
-        if ($existingSettings) {
-            // Create a new instance with existing data and merge new settings
-            $existingData = $existingSettings->toArray();
-            $mergedData = array_merge($existingData, $settings);
-            $accountSettings = new AccountSettings($mergedData);
-        } else {
-            // If no existing settings and no new settings, create empty settings
-            // If there are new settings, use them
-            $accountSettings = new AccountSettings($settings);
-        }
+        
+        // Always merge with existing settings
+        $existingData = $existingSettings->toArray();
+        $mergedData = array_merge($existingData, $settings);
+        $accountSettings = new AccountSettings($mergedData);
 
         $this->pdkSettingsRepository->storeSettings($accountSettings);
 


tests/Unit/Account/Model/AccountGeneralSettingsTest.php
```diff
@@ -0,0 +1,76 @@
...
+namespace MyParcelNL\Pdk\Account\Model;
+
+use MyParcelNL\Pdk\Tests\Uses\UsesMockPdkInstance;
+use function MyParcelNL\Pdk\Tests\usesShared;
+
+usesShared(new UsesMockPdkInstance());
+
+beforeEach(function () {
+    // Clean up any existing cache file
+    $cacheFile = sys_get_temp_dir() . '/pdk_acceptance_api_url.txt';
```
NabDevs, line 14: Only comment would be to potentially convert this to a constant, something like: const ACCEPTANCE_CACHE_FILE = '/pdk_acceptance_api_url.txt';

Other than that, clean PR!

NabDevs (approved):

src/App/Action/Backend/Debug/SwitchToAcceptanceApiAction.php
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * @param  \Symfony\Component\HttpFoundation\Request $request
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function handle(Request $request): Response
+    {
+        try {
+            // Store the acceptance API URL in a file for persistence
+            $acceptanceUrl = 'https://api.acceptance.myparcel.nl';
```
FreekVR, line 54: This should be defined in a constant somewhere

src/App/Action/Backend/Debug/SwitchToAcceptanceApiAction.php
```diff
@@ -0,0 +1,108 @@
...
+            $cacheFile = sys_get_temp_dir() . \MyParcelNL\Pdk\Base\Config::ACCEPTANCE_CACHE_FILE;
+            file_put_contents($cacheFile, $acceptanceUrl);
+
+            // Switch the base URL to the acceptance API for the current session
+            $this->apiService->setBaseUrl('https://api.acceptance.myparcel.nl');
```
FreekVR, line 59: This should be defined in a constant somewhere

FreekVR, line 59: This should also set URLs for microservices (currently direct printing and address services)

src/App/Action/Backend/Debug/SwitchToProductionApiAction.php
```diff
@@ -0,0 +1,107 @@
...
+    {
+        try {
+            // Remove the acceptance API URL from the cache file
+            $cacheFile = sys_get_temp_dir() . \MyParcelNL\Pdk\Base\Config::ACCEPTANCE_CACHE_FILE;
+            if (file_exists($cacheFile)) {
+                unlink($cacheFile);
+            }
+
+            // Switch the base URL back to the production API for the current session
+            $this->apiService->setBaseUrl('https://api.myparcel.nl');
```
FreekVR, line 59: This should come from a constant or config

src/Account/Model/AccountGeneralSettings.php
```diff
@@ -27,4 +27,16 @@ class AccountGeneralSettings extends Model
...
+
+    /**
+     * Get the isTest value, determining it dynamically based on the API environment
+     *
+     * @return bool
+     */
+    public function getIsTestAttribute(): bool
+    {
+        // Check if we're connected to acceptance environment
+        $cacheFile = sys_get_temp_dir() . \MyParcelNL\Pdk\Base\Config::ACCEPTANCE_CACHE_FILE;
```
FreekVR, line 39: This does not seem to be a secure way to track this. Any application may create a file in this directory, causing our application to switch to acceptance without anyone noticing.

FreekVR (changes requested):

src/Account/Model/AccountGeneralSettings.php
joerivanveen: I don’t understand why this is needed if the switch is stored in the account settings (or am I mistaken?).

src/App/Action/Backend/Debug/SwitchToAcceptanceApiAction.php
joerivanveen: I would suggest different logic indeed, that contains a list of easy to maintain url’s to be ‘switched’. Something reordering that probably needs to be done anyway that I neglected to do when building the direct printing option.

joerivanveen (changes requested):

src/Account/Model/AccountGeneralSettings.php
```diff
@@ -7,24 +7,83 @@
...
+
+    /**
+     * Get the current environment
+     *
+     * @return string
+     */
+    public function getEnvironment(): string
+    {
+        // Check if we're connected to acceptance environment via cache file (backward compatibility)
+        $cacheFile = sys_get_temp_dir() . \MyParcelNL\Pdk\Base\Config::ACCEPTANCE_CACHE_FILE;
```
FreekVR, line 39: In its current form, this still contains a security vulnerability where outside applications can influence the `cacheFile` and thus cause side effects in the application to occur.

I would _strongly_ suggest using a different solution e.g. database storage or the `.env`.

src/Account/Model/AccountGeneralSettings.php
```diff
@@ -7,24 +7,83 @@
...
+        if (file_exists($cacheFile)) {
+            return 'acceptance';
```
FreekVR, line 41: These environment names should probably be stored in constant values

src/Api/Service/AbstractApiService.php
```diff
@@ -96,7 +96,27 @@ public function doRequest(
...
+        }
+
+        // Check if we're in acceptance mode via account settings
+        try {
+            $accountSettings = Pdk::get('accountSettings');
+            if ($accountSettings && $accountSettings->isAcceptance()) {
+                return \MyParcelNL\Pdk\Base\Config::API_URL_ACCEPTANCE;
+            }
+        } catch (\Throwable $e) {
+            // Fall back to cache file check for backward compatibility
```
FreekVR, line 111: Can you not call `AccountGeneralSettings->getEnvironment()` here? This contains code duplication now.

src/Api/Service/AbstractApiService.php
```diff
@@ -108,7 +128,21 @@ public function getHeaders(): array
...
 
     /**
-     * @param string $baseUrl
+     * Check if currently connected to acceptance environment
+     * This method is kept for backward compatibility but should be replaced
+     * with checking AccountGeneralSettings->isAcceptance()
+     *
+     * @return bool
+     */
+    public function isConnectedToAcceptance(): bool
```
FreekVR, line 137: This appears to have no actual use outside of the test suite, so I'd suggest removing (and defining this purely in test expectations if needed there)

src/App/Action/Backend/Debug/SwitchToAcceptanceApiAction.php
```diff
@@ -0,0 +1,108 @@
...
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function handle(Request $request): Response
+    {
+        try {
+            // Switch the base URL to the acceptance API for the current session
+            $this->apiService->setBaseUrl(Config::API_URL_ACCEPTANCE);
+
+            // Store the acceptance API URL in a file for persistence (backward compatibility)
+            $acceptanceUrl = Config::API_URL_ACCEPTANCE;
```
FreekVR, line 59: Suggest writing to either a more secure location, or removing this functionality entirely

src/App/Action/Backend/Debug/SwitchToAcceptanceApiAction.php
```diff
@@ -0,0 +1,108 @@
...
+
+            // Call UPDATE_ACCOUNT to reload the context, just like DeleteAccountAction
+            return Actions::execute(PdkBackendActions::UPDATE_ACCOUNT);
+        } catch (Throwable $e) {
+            Logger::error('Failed to switch API URLs to acceptance environment', [
+                'error' => $e->getMessage(),
+                'trace' => $e->getTraceAsString(),
+            ]);
+
+            Notifications::error(
```
FreekVR, line 82: There seems to be some duplication going on here, afaik in standard PDK responses, notifications are always included, so it shouldn't be needed to have both `Notifications:error` and a custom `JsonResponse` here

FreekVR (changes requested):

FreekVR (approved):
Is goed, maar wel even de v3 commits er uit halen, of besluiten het naar v3 te mergen ipv v2

joerivanveen (changes requested):
Just one request, to get this done: can you add the direct printing url as well, preferably in a way that allows for easy adding of other microservices (each on their own url) later?

PR #432, opened 2026-03-31
feat: implement whoami

GravendeelJochem:
INT-1342


src/App/Action/Backend/Account/UpdateSubscriptionFeaturesAction.php
```diff
@@ -4,32 +4,43 @@
...
     /**
-     * @param  \MyParcelNL\Pdk\Account\Repository\AclRepository                   $aclRepository
+     * @param  \MyParcelNL\Pdk\SdkApi\Service\Iam\WhoamiService                   $whoamiService
      * @param  \MyParcelNL\Pdk\App\Account\Contract\PdkAccountRepositoryInterface $pdkAccountRepository
      */
-    public function __construct(AclRepository $aclRepository, PdkAccountRepositoryInterface $pdkAccountRepository)
+    public function __construct(WhoamiService $whoamiService, PdkAccountRepositoryInterface $pdkAccountRepository)
     {
-        $this->aclRepository        = $aclRepository;
+        $this->whoamiService        = $whoamiService;
```
FreekVR, line 43: Is the aclRepository still used now? If its not, we can remove it (merge this into v4-capabilities so we can include it with other breaking changes in case its used by plugins directly)

src/Account/Service/PdkAccountFeaturesService.php
```diff
@@ -0,0 +1,134 @@
...
+     */
+    private function hasFeature(string $featureName): bool
+    {
+        $account = $this->accountRepository->getAccount();
+
+        if (null === $account) {
+            return false;
+        }
+
+        return in_array($featureName, $account->subscriptionFeatures->toArray(), true);
```
FreekVR, line 132: Double check: We're writing different data to subscriptionFeatures (an existing attribute) - do we need to make any updates to its type/model?

FreekVR (changes requested):
A few questions, looks good!

src/App/Action/Backend/Account/UpdateSubscriptionFeaturesAction.php
GravendeelJochem: aclRepository — niet meer aanwezig in de huidige code

src/Account/Service/PdkAccountFeaturesService.php
GravendeelJochem: zelfde type (Collection van strings), geen model-wijzigingen nodig; de SDK enum-deserializer retourneert plain strings net als de oude API-response

src/App/Action/Backend/Order/ExportOrderAction.php
```diff
@@ -102,7 +103,7 @@ public function handle(Request $request): Response
      */
     protected function export(PdkOrderCollection $orders, Request $request): PdkOrderCollection
     {
-        if (! Settings::get(OrderSettings::ORDER_MODE, OrderSettings::ID)) {
+        if (! AccountSettings::usesOrderMode()) {
             return $this->exportShipments($orders);
         }
```
FreekVR, line 109: This could still potentially call the (v1) order API if v2 orders are active. Suggest adding an early return here if Order V2 is active (ensure nothing is exported/return the order collection unchanged). Otherwise we would rely purely on UI to prevent exporting here.

FreekVR (changes requested):
One small addition - looks great otherwise!

NabDevs (approved):

FreekVR (approved):

PR #439, opened 2026-04-03
feat(api): add SDK API infrastructure for generated API clients

FreekVR:
Depends on: #438

Introduce the SdkApi module for calling SDK-generated API clients, replacing
hand-written request/response classes. Includes logging middleware with
sensitive data scrubbing, a SdkBackedModel base class for wrapping SDK types
in PDK-compatible models, and a CapabilitiesService for fetching contract
capability definitions.

- Add SdkApi/ module (AbstractSdkApiService, LoggingMiddleware, CoreApi hierarchy)
- Add CapabilitiesService for fetching contract capability definitions
- Add SdkBackedModel base class and SdkModelHelper utilities
- Add SensitiveDataScrubber for sanitizing log data
- Add test infrastructure (MockSdkApiHandler, UsesSdkApiMock, etc.)
- Fix accept headers in MyParcelApiService

Refs: INT-1261

### DIFF
--- a/src/Base/Model/SdkBackedModel.php
+++ b/src/Base/Model/SdkBackedModel.php
@@ -1,0 +1,259 @@
+<?php
+
+declare(strict_types=1);
+
+namespace MyParcelNL\Pdk\Base\Model;
+
+use MyParcelNL\Pdk\Base\Contract\Arrayable;
+use MyParcelNL\Pdk\Base\Support\SdkModelHelper;
+use MyParcelNL\Pdk\Base\Support\Utils;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\ModelInterface;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\ObjectSerializer;
+
+/**
+ * Base class for PDK Models that are backed by a single SDK generated ModelInterface instance.
+ *
+ * SDK generated models (e.g. from OpenAPI codegen) implement ModelInterface and use a different internal structure
+ * ($container, static getters/setters/attributeMap) than PDK Models ($attributes, HasAttributes trait). This class
+ * bridges the two systems so all SDK model properties are accessible directly on the PDK model.
+ *
+ * Usage:
+ * 1. Extend this class instead of Model.
+ * 2. Declare `protected $sdkModelClass = SomeSdkModel::class;` pointing to the ModelInterface implementation.
+ *    This is the only requirement — the SDK model is discovered and initialised automatically.
+ * 3. Optionally declare native `$attributes` and `$casts` for PDK-only properties on top. Native PDK attributes always take precedence over SDK model properties.
+ */
+abstract class SdkBackedModel extends Model
+{
+    /**
+     * The SDK ModelInterface class this model is backed by.
+     * Subclasses must override this.
+     *
+     * @var string|null
+     */
+    protected $sdkModelClass = null;
+
+    /**
+     * The SDK model instance backing this model.
+     * Lazily initialised on the first write to an SDK property.
+     *
+     * @var object|null
+     */
+    private $sdkModel = null;
+
+    /**
+     * Getter map cached per class: ['camelCaseKey' => 'getterMethodName', ...]
+     *
+     * @var array<string, array<string, string>>
+     */
+    private static $sdkGetterMapCache = [];
+
+    /**
+     * Setter map cached per class: ['camelCaseKey' => 'setterMethodName', ...]
+     *
+     * @var array<string, array<string, string>>
+     */
+    private static $sdkSetterMapCache = [];
+
+    /**
+     * Return the underlying SDK model instance, or null if not yet initialised.
+     *
+     * @return object|null
+     */
+    public function getSdkModel(): ?object
+    {
+        return $this->sdkModel;
+    }
+
+    /**
+     * Pass non-native keys through setAttribute before handing native data to parent.
+     * All routing logic lives in setAttribute — this just ensures fill() sees SDK keys too.
+     *
+     * @param  array|null $data
+     */
+    public function __construct(?array $data = null)
+    {
+        if ($this->sdkModelClass !== null && $data !== null) {
+            $nativeKeys    = array_flip(array_keys($this->attributes));
+            $nonNativeData = array_diff_key($data, $nativeKeys);
+
+            foreach ($nonNativeData as $key => $value) {
+                $this->setAttribute($key, $value);
+            }
+
+            $data = array_intersect_key($data, $nativeKeys);
+        }
+
+        parent::__construct($data);
+    }
+
+    /**
+     * Override getAttribute to proxy unknown keys to the SDK model.
+     * Native PDK attributes take priority.
+     *
+     * @param  string $key
+     *
+     * @return mixed
+     */
+    public function getAttribute(string $key)
+    {
+        $key = Utils::changeCase($key);
+
+        $nativeValue = parent::getAttribute($key);
+
+        if (null !== $nativeValue || array_key_exists($key, $this->attributes)) {
+            return $nativeValue;
+        }
+
+        if ($this->sdkModel !== null) {
+            $getterMap = $this->getSdkGetterMap();
+
+            if (isset($getterMap[$key])) {
+                return $this->sdkModel->{$getterMap[$key]}();
+            }
+        }
+
+        return null;
+    }
+
+    /**
+     * Override setAttribute as the single routing point for all incoming data.
+     *
+     * Routing order:
+     * 1. Native PDK key  → parent::setAttribute
+     * 2. Direct SDK model instance → stored as the backing model
+     * 3. Known SDK setter key → lazy-init backing model if needed, then proxy to its setter
+     * 4. Anything else → parent::setAttribute (dynamic attribute)
+     *
+     * Because Model::fill() calls setAttribute() for every key in $data, no constructor
+     * override is needed — all routing happens here automatically.
+     *
+     * @param  string $key
+     * @param  mixed  $value
+     *
+     * @return self
+     */
+    public function setAttribute(string $key, $value): self
+    {
+        $key = Utils::changeCase($key);
+
+        if (array_key_exists($key, $this->attributes) || $this->isGuarded($key)) {
+            parent::setAttribute($key, $value);
+            return $this;
+        }
+
+        if ($this->sdkModelClass !== null && $value instanceof $this->sdkModelClass) {
+            $this->sdkModel = $value;
+
+            return $this;
+        }
+
+        if ($this->sdkModelClass !== null) {
+            $setterMap = $this->getSdkSetterMap();
+
+            if (isset($setterMap[$key])) {
+                if ($this->sdkModel === null) {
+                    $this->sdkModel = new $this->sdkModelClass();
+                }
+
+                $snakeKey     = SdkModelHelper::toSdkCase($key);
+                $openApiTypes = $this->sdkModelClass::openAPITypes();
+                $type         = $openApiTypes[$snakeKey] ?? null;
+
+                if ($type !== null && ! $this->isAlreadyHydrated($value, $type)) {
+                    $value = ObjectSerializer::deserialize($value, $type);
+                }
+
+                $this->sdkModel->{$setterMap[$key]}($value);
+
+                return $this;
+            }
+        }
+
+        parent::setAttribute($key, $value);
+        return $this;
+    }
+
+    /**
+     * Override attributesToArray to merge SDK model properties at root level.
+     * PDK native attributes take priority over SDK properties.
+     *
+     * @param  null|int $flags
+     *
+     * @return array
+     */
+    public function attributesToArray(?int $flags = null): array
+    {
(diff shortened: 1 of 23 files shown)


src/Base/Model/SdkBackedModel.php
```diff
@@ -0,0 +1,259 @@
...
+     *
+     * @return mixed
+     */
+    public function getAttribute(string $key)
+    {
+        $key = Utils::changeCase($key);
+
+        $nativeValue = parent::getAttribute($key);
+
+        if (null !== $nativeValue || array_key_exists($key, $this->attributes)) {
```
joerivanveen: If I understand this correctly, the attributes are the pdk properties, and the sdk properties live in the sdk model separately. In that case, I think the logic is quite complicated here. Wouldn’t
`if (array_key_exists($key, $this->attributes)) return parent::getAttribute($key);`
accomplish the same? <- sanity check, maybe this is (even) more complicated than I think.
I wish we could forego `array_key_exists` as it is quite costly, but I see why it is needed now.

src/Base/Model/SdkBackedModel.php
```diff
@@ -0,0 +1,259 @@
...
+     *
+     * @return self
+     */
+    public function setAttribute(string $key, $value): self
+    {
+        $key = Utils::changeCase($key);
+
+        if (array_key_exists($key, $this->attributes) || $this->isGuarded($key)) {
+            parent::setAttribute($key, $value);
+            return $this;
```
joerivanveen, line 140: Linter should add an empty line before returns?

src/Base/Model/SdkBackedModel.php
```diff
@@ -0,0 +1,259 @@
...
+                }
+
+                $this->sdkModel->{$setterMap[$key]}($value);
+
+                return $this;
+            }
+        }
+
+        parent::setAttribute($key, $value);
+        return $this;
```
joerivanveen, line 171: Empty line.

src/Base/Model/SdkBackedModel.php
```diff
@@ -0,0 +1,259 @@
...
+
+        if ($this->sdkModelClass !== null) {
+            $setterMap = $this->getSdkSetterMap();
+
+            if (isset($setterMap[$key])) {
+                if ($this->sdkModel === null) {
+                    $this->sdkModel = new $this->sdkModelClass();
+                }
+
+                $snakeKey     = SdkModelHelper::toSdkCase($key);
```
joerivanveen: - `$snakeCase` holds an assumption regarding `::toSdkCase` that it does not control?
- the intermediate var is not really necessary?

src/Base/Model/SdkBackedModel.php
```diff
@@ -0,0 +1,259 @@
...
+
+    /**
+     * Get the getter map for $sdkModelClass, building and caching it on first access.
+     * Uses the class name directly — no instance required.
+     *
+     * @return array<string, string>
+     */
+    private function getSdkGetterMap(): array
+    {
+        if ($this->sdkModelClass === null) {
```
joerivanveen, line 207: Several places do not use yoda style, not a deal breaker for me but I got accustomed to it since MyParcel :-D

src/Base/Model/SdkBackedModel.php
```diff
@@ -0,0 +1,259 @@
...
+     *
+     * @param  mixed  $value
+     * @param  string $type  openAPI type string, e.g. '\Ns\SomeModel' or '\Ns\SomeModel[]'
+     *
+     * @return bool
+     */
+    private function isAlreadyHydrated($value, string $type): bool
+    {
+        // openAPI array types are suffixed with '[]', e.g. '\Ns\SomeModel[]'
+        if (strcasecmp(substr($type, -2), '[]') === 0) {
```
joerivanveen: In what case would this be different than `if ('[]' === substr($type, -2))` <- sanity check

tests/Unit/SdkApi/Middleware/LoggingMiddlewareTest.php
```diff
@@ -0,0 +1,268 @@
...
+function makeLoggingClient(): array
+{
+    $mock  = new MockHandler();
+    $stack = HandlerStack::create($mock);
+    $stack->push(LoggingMiddleware::forApiRequests());
+
+    return [new Client(['handler' => $stack]), $mock];
+}
+
+// forApiRequests() returns a callable
```
joerivanveen:
```suggestion
```

tests/Unit/SdkApi/Service/CoreApi/Shipment/CapabilitiesServiceTest.php
```diff
@@ -0,0 +1,262 @@
...
+        // Expected
+    }
+
+    $errorLogs = $logger->getLogs(LogLevel::ERROR);
+
+    expect($errorLogs)->toHaveCount(1)
+        ->and($errorLogs[0]['message'])->toBe('[PDK]: API request failed')
+        ->and($errorLogs[0]['context'])->toHaveKeys(['error', 'code']);
+});
+
```
joerivanveen, line 206: I don’t think the data scrubber is tested here, this is probably not necessary as we want to keep tests to the point, but could the link break in any way between what this is testing and the data scrubbing?

joerivanveen (approved):

src/Base/Model/SdkBackedModel.php
FreekVR: There is no linter (only in phpstorm configs) so, um, this is gonna be a recurring issue in this code. I'd suggest we skip this until we have a proper code linter here, as I cannot fix these automatically right now

src/Base/Model/SdkBackedModel.php
FreekVR: Yes, the null fallthrough seems a bit redundant. Note that the lookups here are statically cached which should reduce the impact.

tests/Unit/SdkApi/Service/CoreApi/Shipment/CapabilitiesServiceTest.php
FreekVR: No the data scrubber has its own tests. However any changes to the scrubber *could* impact the output of other tests that check the logs. I don't think there's any getting away from that, hopefully it will help devs think about the real-world impact of changes to the scrubber.

src/Base/Model/SdkBackedModel.php
FreekVR: Something to enforce with a linter as well, I think, in the future :)

src/Base/Model/SdkBackedModel.php
FreekVR: Simplified to `if (array_key_exists($key, $this->attributes)) return parent::getAttribute($key);` as suggested. The null fallthrough was indeed redundant.

src/Base/Model/SdkBackedModel.php
FreekVR: Skipping for now — no linter to enforce this automatically. Will address when we set up a proper code style linter.

src/Base/Model/SdkBackedModel.php
FreekVR: Same as above — skipping until we have a linter.

src/Base/Model/SdkBackedModel.php
FreekVR: Good catch on both points. Inlined the variable to remove the intermediate, and renamed `toSdkCase()` → `toOpenApiKey()` to make the contract with `openAPITypes()` explicit rather than assuming snake_case.

src/Base/Model/SdkBackedModel.php
FreekVR: Agreed, something to enforce with a linter in the future. Skipping for now.

src/Base/Model/SdkBackedModel.php
FreekVR: Correct — no difference since `[]` has no alphabetic chars. Simplified to `if ('[]' === substr($type, -2))`.

src/Base/Support/SensitiveDataScrubber.php
```diff
@@ -0,0 +1,193 @@
...
+        'postal',
+        'postal_code',
+        'zip_code',
+        'lat',
+        'lng',
+        'name',
+        'last_name',
+        'lastName',
+        'first_name',
+        'firstName',
```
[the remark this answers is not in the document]
FreekVR: Fixed — lowercased and deduplicated. Since `isSensitiveKey` uses substring matching, patterns like `name` already cover `first_name`/`lastName`/etc, and `zip`/`postal` cover `zip_code`/`postal_code`. Removed all redundant entries and clarified the comment.

src/SdkApi/Service/AbstractSdkApiService.php
```diff
@@ -0,0 +1,159 @@
...
+ * - User agent string construction (platform + PDK version + PHP version)
+ * - Environment detection (production vs acceptance)
+ * - Guzzle client factory with logging middleware
+ *
+ * **Usage:**
+ * Extend this class in namespace-specific abstracts (e.g., AbstractCoreApiService), then create
+ * concrete service classes for specific business domains (e.g., ContractDefinitionsService).
+ *
+ * @see \MyParcelNL\Pdk\SdkApi\Service\CoreApi\CoreApiService
+ * @see \MyParcelNL\Pdk\SdkApi\Service\CoreApi\Capabilities\ContractDefinitionsService
```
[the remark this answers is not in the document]
FreekVR: Fixed — updated to point to the actual classes: `AbstractCoreApiService` and `Shipment\CapabilitiesService`.

src/SdkApi/Service/AbstractSdkApiService.php
```diff
@@ -0,0 +1,159 @@
...
+ */
+abstract class AbstractSdkApiService
+{
+    /**
+     * @return mixed a generated Configuration clas appropiate to the API you're using
+     */
+    abstract function getApiConfig();
```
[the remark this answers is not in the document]
FreekVR: Fixed — corrected typos and added explicit `public` visibility.

tests/Unit/SdkApi/Service/CoreApi/Shipment/CapabilitiesServiceTest.php
```diff
@@ -0,0 +1,262 @@
...
+    // getContractDefinitions goes to /shipments/capabilities/contract-definitions,
+    // which matches the middleware pattern, so let's verify it also gets the header.
+    $service->getContractDefinitions(null);
+
+    expect($service->capturedRequests)->toHaveCount(1);
+
+    $request = $service->capturedRequests[0];
+    // Contract definitions path also contains /shipments/capabilities
+    expect($request->getHeaderLine('Accept'))->toBe('application/json;charset=utf-8;version=2');
+});
```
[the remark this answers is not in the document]
FreekVR: Merged both tests into a single test that verifies the Accept header for all capabilities endpoints (`getCapabilities` and `getContractDefinitions`).

tests/Unit/SdkApi/Service/AbstractSdkApiServiceTest.php
```diff
@@ -0,0 +1,184 @@
...
+it('includes custom user agents from config', function () {
+    $pdk = MockPdkFactory::create([
+        'userAgent' => value([
+            'MyPlatform' => '2.0.0',
+            'CustomApp'  => '1.5.3',
+        ]),
+    ]);
+
+    $service = new ConcreteSdkApiServiceForTest();
+    $userAgent = $service->publicGetUserAgent();
```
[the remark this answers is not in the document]
FreekVR: Removed the variable assignment — `MockPdkFactory::create()` is called for its side effects only.

tests/SdkApi/Response/SdkJsonResponse.php
```diff
@@ -0,0 +1,79 @@
...
+        return ['Content-Type' => 'application/json'];
+    }
+
+    /**
+     * @return int
+     */
+    public function getStatusCode(): int
+    {
+        return SymfonyResponse::HTTP_OK;
+    }
```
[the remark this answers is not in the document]
FreekVR: Removed the no-op `getHeaders()` and `getStatusCode()` overrides. Moved `Content-Type: application/json` to the constructor default for `$headers` instead.

src/SdkApi/Service/CoreApi/Shipment/CapabilitiesService.php
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * **Usage Examples:**
+ * ```php
+ * $service = new CapabilitiesService();
+ *
+ * // Get dynamic capabilities for a specific shipment
+ * $capabilities = $service->getCapabilities([
+ *     'carrier' => 'POSTNL',
+ *     'recipient' => ['cc' => 'NL', 'postal_code' => '2132WT'],
+ *     'package_type' => 'package',
```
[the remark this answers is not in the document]
FreekVR: Fixed — updated to uppercase `PACKAGE` to match SDK enum values.

NabDevs (approved):

PR #443, opened 2026-04-03
refactor!: adapt calculators, delivery options, frontend, and requests to new carrier model

FreekVR:
Depends on: #442

Updates all remaining consumers to work with the refactored carrier model,
capabilities-driven options, and SDK type references. Completes the migration
across calculators, delivery options services, frontend views, context models,
fulfilment/shipment requests, and integration tests.

- Adapt all calculators to new carrier model and capabilities
- Update delivery options services and resources
- Update frontend views and context models
- Update fulfilment and shipment request/response mappings
- Update integration test features, datasets, and fixtures

BREAKING CHANGE: FrontendDataAdapterInterface methods removed:
carrierCollectionToLegacyFormat(), convertCarrierToLegacyFormat(),
getLegacyIdentifier(), getLegacyPackageTypes(), getLegacyDeliveryTypes(),
getLegacyShipmentOptions().
Use getLegacyCarrierIdentifier() instead of getLegacyIdentifier().

DeliveryOptions delivery type and package type ID constants now use SDK enum
values (RefTypesDeliveryType, RefShipmentPackageType) — numeric values may differ.
DeliveryOptions carrier is stored as raw name string, resolved via getter.

CheckoutContext: platformConfig attribute removed.
OrderDataContext: deliveryOptions and inheritedDeliveryOptions now return
arrays in V2 capabilities format instead of DeliveryOptions model instances.

PostShipmentsRequest: carrier ID resolved via CARRIER_NAME_ID_MAP,
contract ID from $shipment->contractId instead of $carrier->contractId.
PostReturnShipmentsRequest: uses CarrierSchema for return capability
validation instead of PropositionService.

Refs: INT-1261


tests/Unit/App/Action/Backend/Order/ExportOrderActionTest.php
```diff
@@ -372,12 +828,12 @@
...
     MockApi::enqueue($errorResponse);
 
-    factory(CarrierSettings::class, Carrier::CARRIER_POSTNL_NAME)->store();
+    factory(CarrierSettings::class, RefCapabilitiesSharedCarrierV2::POSTNL)->store();
     factory(PdkOrder::class)
         ->withExternalIdentifier('error')
         ->withDeliveryOptions(
             factory(DeliveryOptions::class)
-                ->withCarrier(Carrier::CARRIER_DHL_FOR_YOU_NAME)
+                ->withCarrier(RefCapabilitiesSharedCarrierV2::POSTNL)
```
joerivanveen, line 836: Sanity check: how come the carrier changes?

joerivanveen (approved):

tests/Unit/App/Action/Backend/Order/ExportOrderActionTest.php
FreekVR: That was an error in the earlier test. Or perhaps an "intentional error". Anyway, it was creating settings for POSTNL but then creating an order with DHL_FOR_YOU. Here both are correctly PostNL, but since we enqueue an error response, it would always return an error regardless.

NabDevs (approved):

PR #450, opened 2026-04-15
feat(order)!: replace carrier-specific calculators with capabilities API

FreekVR:
Replace hardcoded carrier-specific order calculators with generic
capabilities-driven calculators. Package type, delivery type, option
constraints (requires/excludes/isRequired), insurance tiers, and weight
handling are now derived at runtime from the capabilities API instead of
per-carrier PHP classes and JSON schemas.

## What changed

- New `CapabilitiesPackageTypeCalculator` validates the order's package type
  against shipment-context capabilities (carrier + country + package type)
  and falls back to the next available type when the selected type is not
  supported. Honors the `allowInternationalMailbox` merchant setting.
- New `CapabilitiesDeliveryTypeCalculator` validates the delivery type
  against the carrier's `deliveryTypes` for the resolved package type and
  resets to `standard` (or first available) when the selected type isn't
  supported — replacing the deleted PostNL/DhlForYou/UPS-specific
  delivery-type calculators.
- New `CapabilitiesOptionCalculator` applies API-derived option constraints
  (requires/excludes/isRequired). When the carrier has no capability for
  the shipment context, all defined shipment options are explicitly
  disabled. Merchant `allow*` flags are intentionally NOT consulted at
  order-processing time — they remain a checkout-display concern in
  `DeliveryOptionsService`. This matches main's behavior and prevents
  producing orders the API would reject because capabilities require an
  option the merchant disabled.
- New `DeliveryDateExceptionCalculator` for carriers without delivery-date
  support (Bpost, DPD). Local definition until the API exposes this.
- `InsuranceCalculator` now reads insurance bounds from the per-shipment
  capability (carrier + country + package_type + delivery_type) instead of
  the carrier-global contract definitions. Insurance tiers regain
  fine-grained floor steps (€100, €250, €500) lost when the JSON schemas
  were removed; tier construction is centralized in
  `CapabilitiesValidationService::buildInsuranceTiers` and shared with
  `CarrierSchema` so the carrier-settings UI stays consistent.

## Weight handling

`WeightCalculator` runs late in the chain, so capability checks performed
earlier saw the raw `totalWeight` and could falsely reject otherwise-valid
package types against carrier min-weight constraints. Introduces a shared
`WeightServiceInterface::getEffectiveWeight(physicalProperties, packageType)`
that mirrors `WeightCalculator`'s manualWeight rules and applies the
configured empty-weight fallback per package type. `WeightCalculator` and
`CapabilitiesPackageTypeCalculator` (current type and each fallback
candidate) both delegate to this helper, so the value used for capability
checks matches what the export will eventually submit.

## Capabilities request normalization (V2 wire format)

The PDK runs the V2 capabilities endpoint, but the existing per-call array
shape was half-V1 (`recipient.cc`, raw nested arrays bypassing
`attributeMap`). `SdkApi\…\CapabilitiesService::getCapabilities` now
hydrates nested arrays into the typed SDK models declared in
`openAPITypes` (recipient, sender, physical_properties, options, pickup),
so the wire payload is V2-correct without callers needing to know the wire
keys. Existing `'cc'` callers in PDK code/tests/docblocks were migrated to
`'country_code'`, and regression tests on the SdkApi service pin the wire
JSON to fail loudly if either direction regresses.

## Test infrastructure

- `UsesSdkApiMock` now resets the storage cache around each test, removing
  the per-test `resetCache()` / `resetStorageCache()` helpers that
  individual calculator tests used to call.
- `MockCarrierCapabilitiesRepository` mirrors carrier-model insurance
  bounds in its capability response so existing fixtures using
  `factory(Carrier::class)->withInsurance(...)` keep working after the
  calculator switched to per-shipment bounds.
- New CI summary job `test-unit-summary` provides a matrix-stable required
  status check name (branch rulesets no longer need to list every PHP
  matrix entry individually).

## Depends on

- #449 (`feat/INT-1500-checkout-capabilities-api`)

## Test plan

- [ ] PostNL shipment with age check — signature and only-recipient auto-enabled
- [ ] PostNL morning/evening — required options forced ENABLED even if
      merchant has `allowSignature = false` (capabilities have final say)
- [ ] PostNL pickup outside NL/BE/DE/DK/SE — delivery type resets to standard
- [ ] DHL Europlus shipment — signature always required
- [ ] GLS shipment to EU — signature required, insurance coupled
- [ ] DHL For You shipment abroad — age check, only-recipient, same-day disabled
- [ ] Bpost/DPD shipment — delivery date empty
- [ ] International mailbox without setting — falls back to default package
- [ ] Mailbox order with weight > carrier max — falls back to package
- [ ] Mailbox order with totalWeight=0 and configured emptyMailboxWeight — kept as mailbox (effective weight passes capability min)
- [ ] Insurance for MAILBOX vs PACKAGE — uses per-shipment bounds, not contract-wide
- [ ] Insurance ladder shows €100/€250 floor tiers in carrier settings UI
- [ ] Contract ID on exported shipment — present in API request

## BREAKING CHANGE

- Removed: `CarrierSpecificCalculator`, `AllowedInCarrierCalculator`,
  `PackageTypeCalculator`, `PackageTypeShipmentOptionsCalculator`,
  `TrackedCalculator`, `AbstractCarrierOptionsCalculator`, all per-carrier
  delivery-type/option calculators (PostNL/DhlForYou/UPS/Bpost/etc.),
  `SchemaRepository`, and 44 JSON schema files.
- Calculator chain in `orderCalculators` config changed —
  `CapabilitiesPackageTypeCalculator → TriStateOptionCalculator →
  CapabilitiesDeliveryTypeCalculator → CapabilitiesOptionCalculator`
  replaces the prior carrier-specific chain.
- `CapabilitiesValidationService::capabilitySupportsWeight` /
  `getCapabilitiesForPackageType` removed (callers migrated to direct
  repository calls or the new `buildInsuranceTiers` helper). New
  `WeightServiceInterface::getEffectiveWeight` added for capability-time
  weight resolution.

Refs: INT-1501, INT-930


src/App/Order/Calculator/General/CapabilitiesPackageTypeCalculator.php
```diff
@@ -0,0 +1,138 @@
...
+        $capabilities = $this->capabilitiesService->getCapabilitiesForPackageType($cc, $v2CurrentType);
+        $capability   = $capabilities[$carrier->carrier] ?? null;
+
+        // International mailbox: capabilities determine availability, but the merchant must also enable it.
+        if ($this->isInternationalMailboxBlocked($currentType, $cc, $carrier)) {
+            $capability = null;
+        }
+
+        if ($capability) {
+            return;
```
FreekVR, line 86: Add comment - fall through when capability supported -> no change required

src/App/Order/Calculator/General/DeliveryDateExceptionCalculator.php
```diff
@@ -0,0 +1,26 @@
...
+declare(strict_types=1);
+
+namespace MyParcelNL\Pdk\App\Order\Calculator\General;
+
+use MyParcelNL\Pdk\App\Order\Calculator\AbstractPdkOrderOptionCalculator;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefCapabilitiesSharedCarrierV2;
+
+final class DeliveryDateExceptionCalculator extends AbstractPdkOrderOptionCalculator
+{
+    // @TODO: replace with capabilities check once the API implements deliveryDate
```
FreekVR: turns out this won't be added to capabilities and is considered validation "metadata". So still a TODO for the future, but needs a different api endpiont.

src/Carrier/Service/CapabilitiesValidationService.php
```diff
@@ -53,6 +53,45 @@ public function getCapabilitiesForPackageType(string $cc, string $v2PackageType)
...
+     * @param  string      $v2PackageType  V2 package type
+     * @param  null|string $v2DeliveryType V2 delivery type (optional)
+     *
+     * @return array<string, RefCapabilitiesResponseCapabilityV2>
+     */
+    public function getCapabilitiesForOrderContext(
+        string $carrier,
+        string $cc,
+        string $v2PackageType,
+        ?string $v2DeliveryType = null
```
FreekVR: Given this is specifically mentioning "order context", shouldn't it just then take the order as param? Otherwise, renaming the method to reflect the arguments it is taking and the way it is returning the results.

src/App/Order/Calculator/General/CapabilitiesPackageTypeCalculator.php
```diff
@@ -0,0 +1,139 @@
...
+        $capabilities = $this->capabilitiesService->getCapabilitiesForPackageType($cc, $v2CurrentType);
+        $capability   = $capabilities[$carrier->carrier] ?? null;
+
+        // International mailbox: capabilities determine availability, but the merchant must also enable it.
+        if ($this->isInternationalMailboxBlocked($currentType, $cc, $carrier)) {
+            $capability = null;
+        }
+
+        // Current type supported for this context — no change needed.
+        if ($capability) {
```
NabDevs: This accepts the current package type as soon as the carrier has a matching capability, but I don’t see the order’s actual weight being checked against that capability. The old carrier-specific schemas did validate this, right? Should we validate with `capabilitySupportsWeight()` before keeping the selected package type?

src/App/Order/Calculator/General/CapabilitiesOptionCalculator.php
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    /**
+     * @return void
+     */
+    public function calculate(): void
+    {
+        $capability = $this->getCarrierCapabilities();
+
+        if (! $capability) {
```
NabDevs, line 50: Could you help me check how invalid delivery types are handled now? Here, if the capabilities request for the selected `delivery_type` returns no matching capability, the calculator just returns and the selected delivery type stays unchanged. Is delivery type validation handled somewhere else?

src/App/Order/Calculator/General/CapabilitiesPackageTypeCalculator.php
FreekVR: Refactored to pass carrier to capabilities and then validate weight on our  side (to increase chance of cache hit)

src/App/Order/Calculator/General/CapabilitiesOptionCalculator.php
FreekVR: Good one, I completely missed that - two issues here, unuspported options aren't dropped (no capabilities = all options allowed), no constraints enforced. I'll add a new calculator *and* ensure unsupported options are removed by the calculator.

src/App/Order/Calculator/General/CapabilitiesPackageTypeCalculator.php
```diff
@@ -0,0 +1,163 @@
...
+        $carrier       = $this->order->deliveryOptions->carrier;
+        $cc            = $this->order->shippingAddress->cc;
+        $currentType   = $this->order->deliveryOptions->packageType;
+        $v2CurrentType = DeliveryOptions::PACKAGE_TYPES_V2_MAP[$currentType] ?? null;
+
+        if (! $cc || ! $v2CurrentType) {
+            return;
+        }
+
+        $weight = (int) $this->order->physicalProperties->totalWeight;
```
joerivanveen: Should we not take the empty package weights into account as well, even though package type is tbd at this point, I think ignoring it may cause surprises on export...

joerivanveen (changes requested):

src/App/Order/Calculator/General/CapabilitiesOptionCalculator.php
```diff
@@ -0,0 +1,311 @@
...
+
+            $requiredDef = $definitionsByCapKey[$requiredCapKey] ?? null;
+
+            if (! $requiredDef || ! $requiredDef->getShipmentOptionsKey()) {
+                continue;
+            }
+
+            $this->forceOption($requiredDef->getShipmentOptionsKey(), TriStateService::ENABLED);
+
+            $this->applyRequiresChain($requiredCapKey, $options, $definitionsByCapKey, array_merge($visited, [$requiredCapKey]));
```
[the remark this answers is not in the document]
FreekVR: `main` didnt have any checks for allow* settings on calculator level - only checkout (delivery options display). Restored this and removed the allow check from the calculator instead.

NabDevs (approved):

NabDevs (approved):

joerivanveen (approved):

PR #451, opened 2026-04-17
feat(migration): replace version- with timestamp based migration, record which individual migrations have already ran

FreekVR:
This PR replaces the migraton system with a laravel-style system where individual migrations are tracked as executed rather than assuming this based on a version string. Any migration that has not yet run will be applied, regardless of whether its older or newer than the current release/last migration timestamp. Migration timestamps are only used to determine a logical order in which to run them.

A console command is included to help developers create a new migration stub.

Closes INT-951.

## What this changes

Migrations used to run on a version range (`installedVersion < getVersion() <= currentVersion`). That breaks on release candidates: an RC like `6.3.0-rc.4` compares as _older_ than an installed `6.3.0`, so no migration can run on RC test environments. Additionally, it fixes the problem where developers would have to "predict" a version on which to run the migration, months before any release was scheduled. 

This replaces the version gate with **per-migration identity tracking** and adds a version-independent migration format. Existing versioned migrations keep working unchanged but adding versioned migrations is now deprecated.

## How migrations work now

Two kinds coexist, both tracked by identity in an `applied_migrations` setting — each runs exactly once, when its identity isn't yet recorded:

- **Versioned** (existing, deprecated): a `MigrationInterface` class with `getVersion()`, e.g. `Migration6_1_0`. Identity = class name. Kept only for backwards-compatibility. Continued to be maintained in a manually curated registry in code.
- **Timestamped** (new): a file `YYYY_MM_DD_HHMMSS_<slug>.php` returning an anonymous class extending `AbstractTimestampedMigration`. Identity = filename. This is the new default. Auto-discovered from `migrationDirectory` (default `<rootDir>/src/Migration`; set to `null` to opt out) -- no need to register it.

```php
return new class extends AbstractTimestampedMigration {
    public function up(): void { /* apply */ }
    public function down(): void { /* optional; defaults to no-op */ }
};
```

Within a run, versioned migrations execute first (by semver), then timestamped ones (chronologically by filename).

## Upgrade path

| Situation                           | Behaviour                                                                                                                                                          |
| ----------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
| **Fresh install**                   | All upgrade migrations recorded as applied without running (the install already reflects their end state).                                                         |
| **Upgrade from a pre-tracking PDK** | `applied_migrations` seeded from `installed_version`: versioned migrations ≤ that version are marked applied; timestamped ones are not, so they run retroactively. |
| **Later version bump**              | Only migrations not already applied run.                                                                                                                           |
| **Uninstall**                       | `down()` runs only for migrations that were actually applied.                                                                                                      |

`installed_version` is still written (informational), and existing migrations need no changes.

## Also

`composer console generate:migration <slug>` scaffolds a timestamped stub. This can be run from plugins as soon as their respective PRs land

### DIFF
--- a/docs/superpowers/plans/2026-04-17-per-migration-tracking-and-timestamps.md
+++ b/docs/superpowers/plans/2026-04-17-per-migration-tracking-and-timestamps.md
@@ -1,0 +1,2040 @@
+# Per-Migration Tracking + Timestamp-Based Migrations Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Jira:** [INT-951 — Als PDK plugin wil ik betrouwbaar en vooral eenvoudig migraties kunnen doen](https://myparcelnl.atlassian.net/browse/INT-951). The PR that ships this work must reference INT-951 in the description; commit message templates in this plan include the key.
+
+**Goal:** Replace the version-comparison migration gate with per-migration class/file tracking, and introduce Laravel-style timestamp-based file migrations (anonymous classes) alongside the existing class-based ones. Fix the concrete WooCommerce bug where `Migration6_1_0` is silently skipped on the test RC build because `6.1.0 <= 6.3.0 (installed)` and `6.4.0 > 6.3.0-rc.X`.
+
+**Architecture:** Introduce a new settings option `_myparcelcom_applied_migrations` (a list of migration identities — class FQCN for class-based migrations, filename for file-based migrations). `InstallerService` filters the migration collection by identity instead of by version range. On first run after this change ships, the option is seeded from the legacy `installed_version` so class-based migrations that already ran under the old gate are not re-executed. A new `TimestampedMigrationInterface` + `AbstractTimestampedMigration` base supports Laravel-style migrations where a single file `returns new class extends AbstractTimestampedMigration { … }`. File-based migrations are **never** auto-seeded — they always run once, regardless of when they were added.
+
+**Design rationale for `getId()`:** For class-based migrations, `get_class($migration)` is a stable identity. For anonymous classes (the Laravel-style ones), `get_class()` returns something like `class@anonymous\0/absolute/path/to/file.php:5$42` — not stable across servers/PHP versions. So `getId()` exists only on `TimestampedMigrationInterface` to expose the filename as a portable identity. Class-based migrations use `get_class()` via a fallback in `resolveMigrationId()`. Putting `getId()` on the base `MigrationInterface` would be a breaking change for any external implementer; isolating it in the sub-interface keeps the contract backwards compatible. Since the filename convention is `YYYY_MM_DD_HHMMSS_<slug>.php`, the id string is already chronologically sortable — so sort order uses `strcmp($a->getId(), $b->getId())` and we don't need a separate `getTimestamp()` method (same pattern as Laravel's `Migrator`).
+
+**Tech Stack:**
+- PHP 7.4 (plugin runtime constraint — no constructor property promotion, no readonly properties, no `str_ends_with`; anonymous classes are fine since 7.0)
+- Pest (test framework in both repos)
+- PHP-DI (container used for class-based migration resolution)
+- WordPress options API / PrestaShop settings API (backing storage, abstracted via `PdkSettingsRepositoryInterface`)
+
+**Repositories touched:**
+- `myparcelnl/pdk` (shared framework — primary change)
+- `myparcelnl-woocommerce` (adoption + concrete migration that fixes the carrier bug)
+
+---
+
+## File Structure
+
+### PDK repo (`~/projects/pdk`)
+
+| Path | Responsibility | New? |
+|---|---|---|
+| `src/App/Installer/Contract/MigrationInterface.php` | Base contract for migrations. Unchanged signatures — additive `getId()` is introduced on the timestamped sub-interface so existing implementers don't break. | No change |
+| `src/App/Installer/Contract/TimestampedMigrationInterface.php` | New sub-interface declaring `getId(): string`. | Create |
+| `src/App/Installer/Migration/AbstractTimestampedMigration.php` | Abstract base for file-based migrations. Holds `$id` set by loader. Explicitly disables `getVersion()`. | Create |
+| `src/App/Installer/Service/InstallerService.php` | Core change: new `getAppliedMigrations()`, `markMigrationApplied()`, `resolveMigrationId()`, `discoverTimestampedMigrationFiles()`, file-loader path in `createMigrationCollection()`, identity-based filter replaces version gate in `getUpgradeMigrations()`, updated sort in `runUpMigrations()`, eager seed in `executeInstallation()`. Keeps `installed_version` writes as informational. | Modify |
+| `config/pdk-settings.php` | Add `settingKeyAppliedMigrations` factory and `migrationDirectory` factory (default: `<rootDir>/src/Migration`; plugins override in their own config to change path, or set `null` to disable PDK-owned discovery). | Modify |
+| `tests/Bootstrap/MockTimestampedMigration20260101.php` | Mock fixture for tests. | Create |
+| `tests/Unit/App/Installer/Service/InstallerServiceTest.php` | Add tests for: seeding, identity-based filter, file-loader, timestamp sort, RC-version scenario. | Modify |
+
+### WooCommerce plugin repo (`~/projects/docker-wordpress/plugins/myparcelnl-woocommerce`)
+
+| Path | Responsibility | New? |
+|---|---|---|
+| `src/Pdk/Plugin/Installer/WcMigrationService.php` | **No change required.** PDK-owned discovery finds timestamped migration files automatically. Plugins that want to disable or relocate discovery set `migrationDirectory` in their own config override. | No change |
+| `src/Migration/2026_04_17_100000_migrate_carriers_to_v2.php` | New file-based timestamped migration replicating the body of `Migration6_1_0`. Runs regardless of version, exactly once per install. | Create |
+| `src/Migration/Migration6_1_0.php` | Keep as-is (registered but always seeded as applied for existing installs). Delete once all environments have upgraded past this transition. | No change now; deletion is follow-up |
+| `tests/Unit/Migration/TimestampedCarrierMigrationTest.php` | Test that the new file migration hydrates carriers correctly and is marked applied after running. | Create |
+
+---
+
+## Conventions referenced
+
+- **PDK tests:** run from `~/projects/pdk` with `./vendor/bin/pest <args>`.
+- **WC plugin tests:** must run inside the Docker container (see WC CLAUDE.md). Command pattern:
+
+  ```bash
+  cd ~/projects/docker-wordpress && PHP_VERSION=7.4 docker compose run --rm php bash -c \
+    'cd /var/www/html/wp-content/plugins/myparcelnl-woocommerce && ./vendor/bin/pest <args>'
+  ```
+- **Commits:** conventional commits. Every commit in this plan includes `INT-951` in the footer so Jira links them automatically.
+
+---
+
+## Task 1 (PDK): Add `TimestampedMigrationInterface`
+
+**Files:**
+- Create: `src/App/Installer/Contract/TimestampedMigrationInterface.php`
+
+- [ ] **Step 1: Create the interface file**
+
+```php
+<?php
+
+declare(strict_types=1);
+
+namespace MyParcelNL\Pdk\App\Installer\Contract;
+
+interface TimestampedMigrationInterface extends MigrationInterface
+{
+    /**
+     * Stable unique identifier for this migration. Used for per-migration tracking
+     * AND for ordering — because filenames follow "YYYY_MM_DD_HHMMSS_<slug>", a
+     * lexicographic sort on this string yields chronological order.
+     *
+     * For file-based migrations this is the filename without extension, e.g.
+     * "2026_04_17_100000_migrate_carriers".
+     */
+    public function getId(): string;
+}
+```
+
+- [ ] **Step 2: Commit**
+
+```bash
+cd ~/projects/pdk
+git add src/App/Installer/Contract/TimestampedMigrationInterface.php
+git commit -m "feat(installer): add TimestampedMigrationInterface
+
+INT-951"
+```
+
+---
+
+## Task 2 (PDK): Add `AbstractTimestampedMigration` base class
+
+**Files:**
+- Create: `src/App/Installer/Migration/AbstractTimestampedMigration.php`
+
+- [ ] **Step 1: Create the abstract base**
+
+```php
+<?php
+
+declare(strict_types=1);
+
+namespace MyParcelNL\Pdk\App\Installer\Migration;
+
+use LogicException;
+use MyParcelNL\Pdk\App\Installer\Contract\TimestampedMigrationInterface;
+
+abstract class AbstractTimestampedMigration implements TimestampedMigrationInterface
+{
+    /** @var string */
+    private $id = '';
+
+    /**
+     * Called by the InstallerService loader once the migration file has been required.
+     * Anonymous-class migrations cannot know their own filename, so identity is injected.
+     */
+    public function setIdentity(string $id): void
+    {
+        $this->id = $id;
+    }
+
+    public function getId(): string
+    {
+        if ('' === $this->id) {
+            throw new LogicException('Migration identity has not been set. Ensure the migration is loaded via InstallerService::loadFileMigration().');
+        }
+
+        return $this->id;
+    }
+
+    /**
+     * Timestamp-based migrations are not version-gated.
+     * This method exists solely to satisfy MigrationInterface.
+     */
+    final public function getVersion(): string
+    {
+        throw new LogicException('Timestamp-based migrations are not version-gated. Use getId() for ordering.');
+    }
+
+    public function down(): void
+    {
+        // Default: no-op. Subclasses may override.
+    }
+}
+```
+
+- [ ] **Step 2: Commit**
+
+```bash
+git add src/App/Installer/Migration/AbstractTimestampedMigration.php
+git commit -m "feat(installer): add AbstractTimestampedMigration base class
+
+INT-951"
+```
+
+---
+
+## Task 3 (PDK): Add `settingKeyAppliedMigrations` + `migrationDirectory` config
+
+**Files:**
+- Modify: `config/pdk-settings.php` (append to the returned array, near `settingKeyInstalledVersion`)
+
+- [ ] **Step 1: Add the new factory entries**
+
+Locate the existing `'settingKeyInstalledVersion' => factory(...)` entry near the end of the returned array. Append after it:
+
+```php
+    /**
+     * Settings key where the list of applied migration identities is saved.
+     * Format: string[] of migration ids (FQCN for class-based, filename for file-based).
+     */
+    'settingKeyAppliedMigrations' => factory(function () {
+        return PdkFacade::get('createSettingsKey')('applied_migrations');
+    }),
+
+    /**
+     * Directory the installer scans for timestamped migration files.
+     * Defaults to "<rootDir>/src/Migration". Plugins can override in their
+     * own config to point at a different directory, or set to null to disable
+     * PDK-owned discovery entirely (e.g. if the plugin prefers to register
+     * every source explicitly via its MigrationService).
+     */
+    'migrationDirectory' => factory(function () {
+        return rtrim(PdkFacade::get('rootDir'), '/') . '/src/Migration';
+    }),
+```
+
+- [ ] **Step 2: Commit**
+
+```bash
+git add config/pdk-settings.php
+git commit -m "feat(installer): add settingKeyAppliedMigrations + migrationDirectory config
+
+INT-951"
+```
+
+---
+
+## Task 4 (PDK): Create mock timestamped migration fixture for tests
+
+**Files:**
+- Create: `tests/Bootstrap/MockTimestampedMigration20260101.php`
+
+- [ ] **Step 1: Create the mock**
+
+```php
+<?php
+
+declare(strict_types=1);
+
+namespace MyParcelNL\Pdk\Tests\Bootstrap;
+
+use MyParcelNL\Pdk\App\Installer\Migration\AbstractTimestampedMigration;
+use MyParcelNL\Pdk\Facade\Settings;
+use MyParcelNL\Pdk\Settings\Model\OrderSettings;
+
+final class MockTimestampedMigration20260101 extends AbstractTimestampedMigration
+{
+    public function __construct()
+    {
+        // In production code, setIdentity is called by the InstallerService loader.
+        // The mock is registered as a FQCN, not as a file path, so we self-identify here.
+        $this->setIdentity('2026_01_01_000000_mock_timestamped');
+    }
+
+    public function up(): void
+    {
+        // Write a sentinel so tests can assert this migration ran.
+        Settings::set(sprintf('%s.%s', OrderSettings::ID, 'mockTimestampedMarker'), 'applied');
+
+        if (isset($GLOBALS['__migration_order'])) {
+            $GLOBALS['__migration_order'][] = $this->getId();
+        }
+    }
+}
+```
+
+- [ ] **Step 2: Commit**
+
+```bash
+git add tests/Bootstrap/MockTimestampedMigration20260101.php
+git commit -m "test(installer): add mock timestamped migration fixture
+
+INT-951"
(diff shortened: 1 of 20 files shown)


src/App/Installer/Migration/AbstractTimestampedMigration.php
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * Timestamp-based migrations are not version-gated.
+     * This method exists solely to satisfy MigrationInterface.
+     */
+    final public function getVersion(): string
+    {
+        throw new LogicException('Timestamp-based migrations are not version-gated. Use getId() for ordering.');
+    }
+
+    public function down(): void
```
FreekVR, line 54: why just down, no up?

private/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,120 @@
...
+use RuntimeException;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class MakeMigrationCommand extends Command
+{
+    protected static $defaultName = 'make:migration';
```
FreekVR: Double check: is `make:migration` naming scheme in line with any other (stub) file generators/commands if those pre-exist in the PDK?

private/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,120 @@
...
+
+    private const DEFAULT_UPGRADE_PATH = 'src/Migration';
```
FreekVR: Shouldn't we use the `migrationDirectory`? Migrations need to be created from/in the plugin - not the PDK itself!

private/Command/MakeMigrationCommand.php
```diff
@@ -0,0 +1,120 @@
...
+namespace MyParcelNL\Pdk\Console\Command;
+
+use RuntimeException;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+
+final class MakeMigrationCommand extends Command
```
FreekVR: Document the full signature of the command incl examples (refer to existing commands to see if they already have a standardized docblock for this)

src/App/Installer/Contract/TimestampedMigrationInterface.php
```diff
@@ -0,0 +1,18 @@
...
+
+interface TimestampedMigrationInterface extends MigrationInterface
+{
+    /**
+     * Stable unique identifier for this migration. Used for per-migration tracking
+     * AND for ordering — because filenames follow "YYYY_MM_DD_HHMMSS_<slug>", a
+     * lexicographic sort on this string yields chronological order.
+     *
+     * For file-based migrations this is the filename without extension, e.g.
+     * "2026_04_17_100000_migrate_carriers".
```
FreekVR: Simplify this comment.

- what is "lexicograhic sort"?
- what is a "file-based migration"?
- what is expected to be returned here for non-"file-based migrations"? And what migrations are those anyway?

src/App/Installer/Migration/AbstractTimestampedMigration.php
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Called by the InstallerService loader once the migration file has been required.
+     * Anonymous-class migrations cannot know their own filename, so identity is injected.
+     */
+    public function setIdentity(string $id): void
+    {
+        $this->id = $id;
+    }
+
+    public function getId(): string
```
FreekVR, line 36: Missing docblock, use @inheritdoc at minimum?

src/App/Installer/Service/InstallerService.php
```diff
@@ -92,6 +94,44 @@ protected function executeInstallation(...$args): void
     {
         $this->setDefaultSettings();
         $this->migrateInstall();
+        // migrateInstall() marks the installation migrations as applied; this call then
+        // overwrites applied_migrations with the upgrade-migration ids. That overwrite is
+        // intentional: installation migrations never appear in the upgrade set, so dropping
+        // their ids here cannot cause them to re-run.
```
FreekVR: This comment is a bit unclear. What this does is store all existing migrations in the DB-maintained list (that's the new mechanism that determines what migrations to run on upgrade)

src/App/Installer/Service/InstallerService.php
```diff
@@ -92,6 +94,44 @@ protected function executeInstallation(...$args): void
...
+     * Pre-marks every registered upgrade migration as applied on a fresh install.
+     * This prevents them from firing retroactively on the user's first upgrade —
+     * they're considered "baked into" the installed version.
+     *
+     * Auto-discovered files from migrationDirectory are merged in alongside
+     * plugin-registered sources so that file-only deployments are also pre-marked.
+     *
+     * @return void
+     */
+    protected function seedAppliedMigrationsForFreshInstall(): void
```
FreekVR: Name by intent, not usage.

I'd suggest "markAllMigrationsApplied()"?

src/App/Installer/Service/InstallerService.php
```diff
@@ -112,14 +152,124 @@ protected function getInstalledVersion(): ?string
...
+        if ($migration instanceof TimestampedMigrationInterface) {
+            return $migration->getId();
+        }
+
+        return get_class($migration);
+    }
+
+    /**
+     * Returns the list of migration identities already applied on this installation.
+     * On first access after a plugin upgrade (when the key doesn't exist yet but
```
FreekVR: Clarify this "plugin upgrade" is specifically the upgrade where the changes in this PR are first introduced. Current wording might be interpreted as any plugin upgrade?

src/App/Installer/Service/InstallerService.php
```diff
@@ -112,14 +152,124 @@ protected function getInstalledVersion(): ?string
...
+     * installed_version does), seeds the list from the legacy installed_version
+     * gate so class-based migrations that already ran are not re-executed.
+     *
+     * Fresh installs bypass this lazy seed path — executeInstallation() seeds
+     * eagerly with ALL registered upgrade migrations. See that method for why.
+     *
+     * @param  null|\MyParcelNL\Pdk\Base\Support\Collection $allMigrations Required on first access to compute the seed.
```
FreekVR: Why doesn't the method self-load the $allMigrations collection? Leaving this to the caller depending on context seems fragile

src/App/Installer/Service/InstallerService.php
```diff
@@ -112,14 +152,124 @@ protected function getInstalledVersion(): ?string
...
+        // returns an empty array for a key that was never stored, so it cannot be
+        // distinguished from a deliberately-empty list. Re-seeding an empty list is
+        // idempotent, so treating empty as absent is safe.
+        if (is_array($stored) && ! empty($stored)) {
+            return $stored;
+        }
+
+        if (null === $allMigrations) {
+            // Cannot seed without the collection; return empty but do NOT persist.
+            return [];
```
FreekVR: I guess my previous comment was on the nose, because here we have a silent fallthrough that doesnt alert the developer to any problem

src/App/Installer/Service/InstallerService.php
```diff
@@ -112,14 +152,124 @@ protected function getInstalledVersion(): ?string
...
-                ->filter(function (MigrationInterface $migration) {
+                ->filter(function (MigrationInterface $migration) use ($applied) {
+                    // Timestamp-based migrations are version-less, so they can't be
+                    // version-gated. Only reverse one that was actually applied — otherwise a
+                    // migration whose up() never ran on this install would have its down() run.
+                    if ($migration instanceof TimestampedMigrationInterface) {
+                        return in_array($this->resolveMigrationId($migration), $applied, true);
+                    }
+
                     return version_compare($migration->getVersion(), $this->getInstalledVersion(), '<=');
```
FreekVR: Given the upgrade path in `getAppliedMigrations` will ALSO seed any version-gated "old style" migrations to the applied migrations table, wouldn't it be better to also use that mechanism here for those types of migrations? Avoids having the same `version_compare` checks in several places.

src/App/Installer/Service/InstallerService.php
```diff
@@ -212,22 +414,67 @@ private function getUpgradeMigrations(?string $version = null): Collection
...
             : $this->migrationService->getUpgradeMigrations();
 
-        $collection = $this->createMigrationCollection($migrations);
+        // Merge PDK-owned file discovery on top, then dedupe. A source can
+        // legitimately appear in both (plugin explicitly registered a file
+        // that's also in migrationDirectory) — dedupe prevents double-require.
+        // Dedupe compares source strings, so a file registered under a different
+        // path than discovery returns (e.g. via a symlink) would not be deduped;
+        // plugins should rely on discovery alone or register the exact discovered path.
+        $sources = array_values(array_unique(array_merge(
```
FreekVR: Avoid chained array methods, unclear what this is doing exactly? Perhaps `Arr::` helpers could be applied here also?

src/App/Installer/Service/InstallerService.php
```diff
@@ -112,14 +152,124 @@ protected function getInstalledVersion(): ?string
...
+    }
+
     /**
      * @return void
      */
     protected function migrateDown(): void
     {
+        $applied = $this->getAppliedMigrations();
+
         $this->runDownMigrations(
```
FreekVR, line 259: Double check: are down migrations ran in the exact reverse order of the up migrations?

src/App/Installer/Service/InstallerService.php
```diff
@@ -212,22 +414,67 @@ private function getUpgradeMigrations(?string $version = null): Collection
...
 
         if (! $version) {
             return $collection;
         }
 
-        return $collection->filter(function (MigrationInterface $migration) use ($version) {
-            return version_compare($migration->getVersion(), $this->getInstalledVersion(), '>')
-                && version_compare($migration->getVersion(), $version, '<=');
+        // Trigger seeding on first access (pass collection so seed can be computed).
+        $applied = $this->getAppliedMigrations($collection);
```
FreekVR: Again, either split the seeding and getter concerns, or skip the optional-specific-usecase param

FreekVR (commented):
Several concerns. One major missing point seems to be incomplete test coverage, especially around (timestamp) migration identification and ordering?

src/App/Installer/Migration/AbstractTimestampedMigration.php
FreekVR: up() is left abstract on purpose — it's the actual migration work, so every concrete migration has to implement it; a default would let one silently do nothing. down() defaults to a no-op since most data migrations are one-way. Added a class docblock explaining that.

private/Command/MakeMigrationCommand.php
FreekVR: Renamed to `generate:migration` (and the class to `GenerateMigrationCommand`) to line up with the existing `generate:*` commands.

private/Command/MakeMigrationCommand.php
FreekVR: Looked into this in depth. The command runs in the standalone console without the plugin's container, so it can't read the plugin's runtime `migrationDirectory`; and the PDK has no `src/Migration` of its own, so it can't scaffold into itself either. It writes to the cwd's `src/Migration`, which matches the default `migrationDirectory` — I verified it works from the actual WooCommerce checkout under `pdk-dev-on`, creating the file in the plugin. Custom dirs go via `--dir`. Also added a `composer console` script to both plugins (separate PRs, Refs INT-951) so it's actually runnable from the plugin context.

private/Command/MakeMigrationCommand.php
FreekVR: Added a `setHelp()` block, so `composer console help generate:migration` now shows the full usage, examples and the `--dir` option.

src/App/Installer/Service/InstallerService.php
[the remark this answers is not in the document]
FreekVR: Good catch. migrateDown now reverses migrations by their recorded applied state instead of version_compare, so an applied migration isn't skipped when installed_version is a pre-release. Added a regression test for exactly that RC scenario.

src/App/Installer/Contract/TimestampedMigrationInterface.php
FreekVR: Simplified — dropped "lexicographic" and "file-based migration". It now just says the id is the filename without extension, used to track the migration and, because it's timestamp-prefixed, to order it. This interface is only implemented by timestamped migrations, so there's no other case to cover.

src/App/Installer/Migration/AbstractTimestampedMigration.php
FreekVR: Added `@inheritDoc`.

src/App/Installer/Service/InstallerService.php
FreekVR: Reworded — it now says it records every registered/discovered upgrade migration as applied so none of them re-run on the next upgrade.

src/App/Installer/Service/InstallerService.php
FreekVR: Renamed to `markAllUpgradeMigrationsApplied()`.

src/App/Installer/Service/InstallerService.php
FreekVR: Clarified in the docblock — it's specifically the upgrade where this tracking is first introduced (applied list missing but installed_version present), not any upgrade.

src/App/Installer/Service/InstallerService.php
FreekVR: Agreed it was fragile. Split it into a pure `getAppliedMigrations()` getter and an explicit `seedAppliedMigrationsFromInstalledVersion(Collection)` seeder, so there's no caller-dependent optional param anymore.

src/App/Installer/Service/InstallerService.php
FreekVR: This goes away with the split above — the getter just returns the stored list (or empty), and the seeder's early-returns are explicit and documented now.

src/App/Installer/Service/InstallerService.php
FreekVR: Done — migrateDown filters by applied_migrations for both versioned and timestamped now, so the version_compare isn't duplicated here. Same change that fixes the RC issue Copilot flagged.

src/App/Installer/Service/InstallerService.php
FreekVR: Pulled it out into a named `mergeMigrationSources()` helper (it was duplicated in the fresh-install seeder too, so that's deduped now).

src/App/Installer/Service/InstallerService.php
FreekVR: Yep — `runDownMigrations` sorts with `compareMigrations($b, $a)`, the exact reverse of the up order (timestamped first, then versions descending). Added a test pinning it.

src/App/Installer/Service/InstallerService.php
FreekVR: Split — `seedAppliedMigrationsFromInstalledVersion()` does the seeding and `getAppliedMigrations()` is now just the getter, so the optional param is gone.

src/App/Installer/Service/InstallerService.php
```diff
@@ -258,11 +536,10 @@ private function runDownMigrations(Collection $migrations): void
...
     {
         $migrations
-            ->sort(function (MigrationInterface $a, MigrationInterface $b) {
-                return version_compare($a->getVersion(), $b->getVersion());
-            })
+            ->sort([$this, 'compareMigrations'])
             ->each(function (MigrationInterface $migration) {
                 $migration->up();
+                $this->markMigrationApplied($migration);
             });
```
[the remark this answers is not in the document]
FreekVR: Added a guard in `runUpMigrations()` — it now runs each migration identity at most once per pass, so two collection entries resolving to the same migration (e.g. divergent path strings that the string-based source dedupe misses) can't both run `up()`. Added a divergent-path test that asserts it runs once.

tests/Unit/App/Installer/Service/InstallerServiceTest.php
```diff
@@ -242,3 +246,540 @@ function expectSettingsToContain(array $values): void
...
+
+        $applied = $settingsRepository->get($appliedMigrationsKey);
+
+        expect($applied)->toContain('2026_06_01_000000_autodiscover_test');
+        expect($settingsRepository->get('order.autoDiscoverMarker'))->toBe('applied');
+    } finally {
+        Pdk::set('migrationDirectory', null);
+        @unlink($file);
+        @rmdir($tmpDir);
+    }
```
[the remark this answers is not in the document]
FreekVR: Good point. It now captures the original `migrationDirectory` up front and restores that in `finally`, instead of forcing `null`.

tests/Unit/App/Installer/Service/InstallerServiceTest.php
```diff
@@ -242,3 +246,540 @@ function expectSettingsToContain(array $values): void
...
+    try {
+        Installer::install();
+
+        expect($GLOBALS['__dedupe_runs'])->toBe(1);
+    } finally {
+        Pdk::set('migrationDirectory', null);
+        unset($GLOBALS['__dedupe_runs']);
+        @unlink($file);
+        @rmdir($tmpDir);
+    }
```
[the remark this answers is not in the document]
FreekVR: Same fix here — captures and restores the original `migrationDirectory` value rather than setting `null`.

tests/Unit/Console/Command/GenerateMigrationCommandTest.php
```diff
@@ -0,0 +1,142 @@
+<?php
+/** @noinspection StaticClosureCanBeUsedInspection */
+
+declare(strict_types=1);
+
+namespace MyParcelNL\Pdk\Console\Command;
+
+use MyParcelNL\Pdk\Console\Command\GenerateMigrationCommand;
+use Symfony\Component\Console\Tester\CommandTester;
+
```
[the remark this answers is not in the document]
FreekVR: Fixed the flakiness — added a namespaced `date()` override so the generated timestamp is deterministic, and rewrote the collision test to pin it (`__fake_now`), so there's no same-second race between the two runs.

tests/Bootstrap/MockMigrationService.php
```diff
@@ -8,6 +8,54 @@
...
+     * Appends a migration class name to the dynamic extras list for the current test.
+     *
+     * @param  string $source
+     *
+     * @return void
+     */
+    public static function addUpgradeMigration(string $source): void
+    {
+        self::$extraUpgrades[] = $source;
+    }
```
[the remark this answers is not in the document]
FreekVR: Updated the docblocks — they now state the entries are class names or absolute file paths.

src/App/Installer/Contract/UpgradeMigrationInterface.php
```diff
@@ -6,6 +6,13 @@
...
 /**
  * Migrations that are run during an upgrade.
+ *
+ * @deprecated Create a timestamped migration instead: a YYYY_MM_DD_HHMMSS_<slug>.php file
+ *   returning an anonymous class that extends
+ *   {@see \MyParcelNL\Pdk\App\Installer\Migration\AbstractTimestampedMigration}.
+ *   Version-based upgrade migrations do not run reliably on release-candidate builds —
+ *   version_compare treats an RC as older than its release, so the migration never fires.
+ *   See INT-951.
  */
```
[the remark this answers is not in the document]
FreekVR: Reworded. You're right that the RC framing was wrong — identity tracking already makes versioned migrations run on RC builds, so that's no longer a reason. The actual reason to prefer timestamped: a version-based migration forces predicting the release up front, which is usually stale by the time the feature ships, so the version now only affects ordering. The docblock says that now.

docs/superpowers/plans/2026-04-17-per-migration-tracking-and-timestamps.md
```diff
@@ -0,0 +1,2077 @@
...
+# → creates ./src/Migration/2026_04_17_143052_migrate_carriers_to_v2.php
+
+# Override the output dir:
+composer console make:migration migrate_carriers_to_v2 --upgrade-path=src/CustomMigrations
+```
+
+**Files:**
+- Create: `private/Command/MakeMigrationCommand.php`
+- Modify: `private/PdkConsoleApp.php` (register the new command)
+- Create: `tests/Unit/Console/Command/MakeMigrationCommandTest.php`
```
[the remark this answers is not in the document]
FreekVR: I left this one mostly as-is. It's a historical planning document rather than living docs, and it already deviates from the shipped design in several places (the seeding/getter split, naming, etc.). I added a banner at the top flagging it as historical and pointing to the code, the README and the Confluence page for the final design (`generate:migration`), rather than syncing it line by line.

private/Command/GenerateMigrationCommand.php
```diff
@@ -0,0 +1,143 @@
...
+            return 1;
+        }
+
+        $written = file_put_contents($path, $this->renderStub());
+
+        if (false === $written) {
+            throw new RuntimeException(sprintf('Failed to write migration to %s', $path));
+        }
+
+        $output->writeln(sprintf('<info>Created: %s</info>', $path));
```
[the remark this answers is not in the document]
FreekVR: this is HIGHLY unlikely. Ignoring

src/App/Installer/Service/InstallerService.php
```diff
@@ -243,26 +495,67 @@ private function runDownMigrations(Collection $migrations): void
...
+     * semantic version and run before timestamp-based ones. Timestamp-based
+     * migrations are ordered lexicographically by their ID (YYYY_MM_DD_HHMMSS_<slug>),
+     * which is equivalent to chronological order.
+     *
+     * @param  \MyParcelNL\Pdk\App\Installer\Contract\MigrationInterface $a
+     * @param  \MyParcelNL\Pdk\App\Installer\Contract\MigrationInterface $b
+     *
+     * @return int
+     */
+    public function compareMigrations(MigrationInterface $a, MigrationInterface $b): int
```
GravendeelJochem: Is er een reden dat compareMigrations() public is? Het lijkt alleen intern gebruikt te worden als sorteringscomparator in runDownMigrations.

GravendeelJochem (approved):
Kleine vraag voor de rest ✅

src/App/Installer/Service/InstallerService.php
FreekVR: Good point, it's purely internal. The only reason it was public is that runUpMigrations passed it straight to Collection::sort() as [$this, 'compareMigrations'], and sort() runs uasort from inside the Collection class — a private method there throws "cannot access private method". So I made it private and wrapped that call in a closure like runDownMigrations already does, so the visibility now matches the usage. Done in acd32245.

docs/superpowers/plans/2026-04-17-per-migration-tracking-and-timestamps.md
```diff
@@ -0,0 +1,2101 @@
+# Per-Migration Tracking + Timestamp-Based Migrations Implementation Plan
+
+> **Historical planning document.** This is the original plan; the shipped implementation deviated from it during review (e.g. the console command is `generate:migration` / `GenerateMigrationCommand`, not `make:migration`, and the seeding/getter were split). For the final design see the code, the PDK `README.md`, and the [Confluence page](https://myparcelnl.atlassian.net/wiki/spaces/IG/pages/866943327). Don't treat the details below as current.
```
joerivanveen, line 3: Should we not remove (or not commit) this superpower if no longer needed?

private/Command/GenerateMigrationCommand.php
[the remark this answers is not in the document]
joerivanveen: Can we reconsider? I have seen the same migration running multiple concurrent times, depending on platform / requests / busyness of the website. Why not use an atomic method if not too much trouble?

joerivanveen (changes requested):

docs/superpowers/plans/2026-04-17-per-migration-tracking-and-timestamps.md
FreekVR: I don't know what the best practice should be? The pro is that agents and developers can refer to the original intent/explainer if they want to make changes. The con is that nobody is likely to clean it up and it may just be stale noise at some point :/

private/Command/GenerateMigrationCommand.php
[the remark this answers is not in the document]
FreekVR: This is purely a local-only command to create a stub migration file that would later be versioned into git. This would NEVER be called in production - in fact it doesn't even work if the dev-version of the PDK isnt linked - and the directory shouldn't be writable to PHP anyway.

joerivanveen (approved):

PR #453, opened 2026-04-17
feat(capabilities): allow calling capabilities proxy from admin, fix issues with camelCase keys, allow filtering by supported options

FreekVR:
**Story:** [INT-930](https://myparcelnl.atlassian.net/browse/INT-930) — Capabilities API migration
**Subtask:** [INT-1505](https://myparcelnl.atlassian.net/browse/INT-1505)

## Summary

Extend the existing `CapabilitiesAction` (introduced by #408) so the JS-PDK admin order-edit form can drive contextual capability lookups through the shared proxy:

- Opt-in `filterOptions: true` body flag applies the registered-options allowlist (`Carrier::filterRegisteredOptions`) to each capability's `options`. Default behavior stays an unfiltered 1:1 SDK passthrough so the delivery-options widget caller is unaffected.
- Server-side PDK-camelCase → SDK-snake_case mapping for `packageType` and `deliveryType`, so JS callers don't need to know SDK conventions. Snake_case inputs continue to work unchanged.

JS-PDK admin wiring (query composable + watcher) ships in a follow-up PR.

## Plan

[\`docs/superpowers/plans/2026-05-01-capabilities-action-v3.md\`](https://github.com/myparcelnl/pdk/blob/feat/INT-1505-admin-capabilities-context/docs/superpowers/plans/2026-05-01-capabilities-action-v3.md)

## Test plan
- [x] PHP 7.4 + 8.4 unit tests pass (`CapabilitiesActionTest`, 7/7)
- [x] No PHPStan regressions
- [ ] Manual smoke: existing delivery-options widget still receives unfiltered options
- [ ] Manual smoke: action with \`filterOptions: true\` returns only registered keys

## Depends on
#408 (already merged into \`v4-capabilities\`)

[INT-930]: https://myparcelnl.atlassian.net/browse/INT-930?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[INT-1505]: https://myparcelnl.atlassian.net/browse/INT-1505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/App/Action/Capabilities/CapabilitiesAction.php
+++ b/src/App/Action/Capabilities/CapabilitiesAction.php
@@ -70,7 +70,10 @@ class CapabilitiesAction implements ActionInterface
      */
     private function prepareSdkArgs(array $payload): array
     {
-        $shouldFilterOptions = ! empty($payload['filterOptions']);
+        $filterOptions        = $payload['filterOptions'] ?? false;
+        $shouldFilterOptions  = is_bool($filterOptions)
+            ? $filterOptions
+            : filter_var($filterOptions, FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE) ?? false;
         unset($payload['filterOptions']);
 
         $camelToSnake = array_flip(CapabilitiesPostCapabilitiesRequestV2::attributeMap());
--- a/src/App/Action/Capabilities/CapabilitiesAction.php
+++ b/src/App/Action/Capabilities/CapabilitiesAction.php
@@ -95,8 +95,9 @@ class CapabilitiesAction implements ActionInterface
         $body = json_decode($request->getContent(), true);
 
         if (is_array($body)) {
-            $wrapped  = $body['data']['capabilities'] ?? null;
-            $bodyData = is_array($wrapped) ? $wrapped : ($body['data'] ?? $body);
+            $data    = $body['data'] ?? null;
+            $wrapped = is_array($data) ? ($data['capabilities'] ?? null) : null;
+            $bodyData = is_array($wrapped) ? $wrapped : (is_array($data) ? $data : $body);
         } else {
             $bodyData = [];
         }
--- a/src/App/Action/Capabilities/CapabilitiesAction.php
+++ b/src/App/Action/Capabilities/CapabilitiesAction.php
@@ -6,8 +6,10 @@ namespace MyParcelNL\Pdk\App\Action\Capabilities;
 
 use MyParcelNL\Pdk\Api\Handler\CorsHandler;
 use MyParcelNL\Pdk\App\Action\Contract\ActionInterface;
+use MyParcelNL\Pdk\Carrier\Model\Carrier;
 use MyParcelNL\Pdk\Facade\Pdk;
 use MyParcelNL\Pdk\SdkApi\Service\CoreApi\Shipment\CapabilitiesService;
+use MyParcelNL\Pdk\Shipment\Model\DeliveryOptions;
 use MyParcelNL\Sdk\Client\Generated\CoreApi\ApiException as CoreApiException;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -41,16 +43,45 @@ class CapabilitiesAction implements ActionInterface
             return $corsHandler->handlePreflightRequest($request) ?? new Response();
         }
 
-        $payload = $this->getRequestData($request);
+        $payload                                = $this->getRequestData($request);
+        [$sdkArgs, $shouldFilterOptions]        = $this->prepareSdkArgs($payload);
 
         try {
-            $results  = $this->capabilitiesService->getCapabilities($payload);
-            $response = $this->createSymfonyResponse($results);
+            $results  = $this->capabilitiesService->getCapabilities($sdkArgs);
+            $response = $this->createSymfonyResponse($results, $shouldFilterOptions);
         } catch (Throwable $e) {
             $response = $this->createErrorResponse($e);
         }
 
         return $corsHandler->addCorsHeaders($request, $response);
+    }
+
+    /**
+     * Pull the action's control parameter ($filterOptions) out of the payload and translate any
+     * PDK-camelCase argument names to SDK-snake_case so callers don't have to know SDK conventions.
+     *
+     * @param  array $payload
+     *
+     * @return array{0: array, 1: bool} [SDK args (control flag stripped), shouldFilterOptions]
+     */
+    private function prepareSdkArgs(array $payload): array
+    {
+        $shouldFilterOptions = ! empty($payload['filterOptions']);
+        unset($payload['filterOptions']);
+
+        if (isset($payload['packageType'])) {
+            $payload['package_type'] = DeliveryOptions::PACKAGE_TYPES_V2_MAP[$payload['packageType']]
+                ?? $payload['packageType'];
+            unset($payload['packageType']);
+        }
+
+        if (isset($payload['deliveryType'])) {
+            $payload['delivery_type'] = DeliveryOptions::DELIVERY_TYPES_V2_MAP[$payload['deliveryType']]
+                ?? $payload['deliveryType'];
+            unset($payload['deliveryType']);
+        }
+
+        return [$payload, $shouldFilterOptions];
     }
 
     private function getRequestData(Request $request): array
@@ -66,11 +97,25 @@ class CapabilitiesAction implements ActionInterface
 
     /**
      * @param  \MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefCapabilitiesResponseCapabilityV2[] $results
+     * @param  bool                                                                                $filterOptions
+     *                                                                                                Apply the registered-options allowlist to each capability's options when true. Default false
+     *                                                                                                preserves the unfiltered SDK passthrough for existing callers.
      */
-    private function createSymfonyResponse(array $results): Response
+    private function createSymfonyResponse(array $results, bool $filterOptions = false): Response
     {
+        $body = json_decode(json_encode(['results' => $results]), true);
+
+        if ($filterOptions && isset($body['results']) && is_array($body['results'])) {
+            $body['results'] = array_map(static function (array $capability): array {
+                if (isset($capability['options']) && is_array($capability['options'])) {
+                    $capability['options'] = Carrier::filterRegisteredOptions($capability['options']);
+                }
+                return $capability;
+            }, $body['results']);
+        }
+
         return new Response(
-            json_encode(['results' => $results]),
+            json_encode($body),
             Response::HTTP_OK,
             ['Content-Type' => 'application/json']
         );
--- a/src/App/Action/Capabilities/CapabilitiesAction.php
+++ b/src/App/Action/Capabilities/CapabilitiesAction.php
@@ -9,8 +9,8 @@ use MyParcelNL\Pdk\App\Action\Contract\ActionInterface;
 use MyParcelNL\Pdk\Carrier\Model\Carrier;
 use MyParcelNL\Pdk\Facade\Pdk;
 use MyParcelNL\Pdk\SdkApi\Service\CoreApi\Shipment\CapabilitiesService;
-use MyParcelNL\Pdk\Shipment\Model\DeliveryOptions;
 use MyParcelNL\Sdk\Client\Generated\CoreApi\ApiException as CoreApiException;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\CapabilitiesPostCapabilitiesRequestV2;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Throwable;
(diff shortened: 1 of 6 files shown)


src/App/Action/Shared/Context/FetchContextAction.php
```diff
@@ -29,11 +30,39 @@ public function __construct(ContextServiceInterface $contextService)
...
+     * Tolerant on shape: missing/empty body and missing 'order' key both return null so the
+     * action stays GET-compatible for callers that don't need order projection.
+     */
+    private function readOrderFromRequest(Request $request): ?PdkOrder
+    {
+        $rawBody = $request->getContent();
+
+        if (! is_string($rawBody) || $rawBody === '') {
+            return null;
+        }
```
FreekVR: Would a simple empty check here not suffice?

src/Carrier/Model/Carrier.php
```diff
@@ -166,6 +166,14 @@ class Carrier extends SdkBackedModel
      */
     protected $casts = [];
 
+    /**
+     * Per-option requires/excludes lists, keyed by capability option key. Populated only when
+     * the carrier has been projected for an order context via withOptionRelationships().
+     *
+     * @var array<string, array{requires: string[], excludes: string[]}>
+     */
+    private $optionRelationships = [];
```
FreekVR: Tracking this here doesn't make sense. Excludes/Requires should live on each individual option instead.

Note: There's a difference between Options tracked on the Carrier (which is based on contract definitions) and the Options/Carrier data returned by the capabilities (sans contract-definitions subpath). Only the specific capabilities contain the requires/exclude data. Avoid polluting models with utility data and try to keep the shape aligned with the API.

src/Carrier/Model/Carrier.php
```diff
@@ -267,4 +287,92 @@ public function getOptionMetadata(string $capabilitiesKey)
...
+     * packageTypes and deliveryTypes to those present in the given results.
+     *
+     * Empty $capabilities returns a clone unchanged (e.g. draft order without
+     * shipping context — narrowing would be lossy without input data).
+     *
+     * @param  \MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefCapabilitiesResponseCapabilityV2[] $capabilities
+     *
+     * @return self
+     */
+    public function withContextualCapabilities(array $capabilities): self
```
FreekVR: I feel like this doesn't belong on the model but either on a repository or capabilities-specific service?

src/Carrier/Model/Carrier.php
```diff
@@ -267,4 +287,92 @@ public function getOptionMetadata(string $capabilitiesKey)
...
+            foreach ($capability->getPackageTypes() ?? [] as $packageType) { // @phpstan-ignore-line SDK declares non-nullable but runtime may be null
+                $allowedPackageTypes[$packageType] = true; // @phpstan-ignore-line SDK declares enum type but returns string
+            }
+            foreach ($capability->getDeliveryTypes() ?? [] as $deliveryType) { // @phpstan-ignore-line SDK declares non-nullable but runtime may be null
+                $allowedDeliveryTypes[$deliveryType] = true; // @phpstan-ignore-line SDK declares enum type but returns string
+            }
+        }
+
+        $clone->packageTypes  = array_values(array_intersect($this->packageTypes ?? [], array_keys($allowedPackageTypes)));
+        $clone->deliveryTypes = array_values(array_intersect($this->deliveryTypes ?? [], array_keys($allowedDeliveryTypes)));
```
FreekVR: Intersecting is unneeded. It's possible to operate purely on the (specific) capabilities response. It can never return *more* package types than the contract definition. It will only narrow - so no need to compare. Treat data as 100% reliable.

src/Carrier/Model/Carrier.php
```diff
@@ -267,4 +287,92 @@ public function getOptionMetadata(string $capabilitiesKey)
...
+     * response, not on the static option definitions — so they only become visible when the
+     * carrier is projected for an order context.
+     *
+     * If multiple capabilities carry data for the same option key, the first one wins.
+     *
+     * @param  \MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefCapabilitiesResponseCapabilityV2[] $capabilities
+     *
+     * @return self
+     */
+    public function withOptionRelationships(array $capabilities): self
```
FreekVR: SImilar concerns as `withContextualCapabilities`. I dont feel like this is the right place to (A) intepret this data and (B) track/store it.

src/Carrier/Service/CapabilitiesValidationService.php
```diff
@@ -56,6 +57,47 @@ public function getCapabilitiesForPackageType(string $cc, string $v2PackageType)
...
+     *
+     * @return RefCapabilitiesResponseCapabilityV2[]
+     */
+    public function getContextualCarrierCapabilities(Carrier $carrier, array $context): array
+    {
+        if (empty($context['cc']) || empty($carrier->carrier)) {
+            return [];
+        }
+
+        $packageTypes = $context['packageTypes'] ?? $carrier->packageTypes ?? [];
```
FreekVR: An earlier PR we introduced some logic to drill down an order into one specific package type. My concern here is that having multiple package types' capabilities combined introduces some complexity and possibly divergent business logic between admin/checkout. Is it feasible to resolve the package type to a single one - using the same existing logic - before calling capabilities?

src/Context/Model/DynamicContext.php
```diff
@@ -70,6 +78,72 @@ public function __construct(?array $data = null)
         $this->attributes['account']  = AccountSettings::getAccount();
         $this->attributes['carriers'] = AccountSettings::getCarriers();
         $this->attributes['shop']     = AccountSettings::getShop();
+
+        if ($order) {
+            $this->projectCarriersForOrder($order);
```
FreekVR: Rather than projecting all the carriers up front - I think its better to alter the approach: pass in as much contextual data from the JS-PDK side, including selected carriers/package types/options/delivery type etc. and make it re-fetch the context on every change within the orders delivery options. That way we tackle two issues:

- Having to fetch all possible combinations up front
- Having complex business logic on the PHP or JS-PDK side for filtering out the correct capabilities to apply given what's being selected

I.o.w. I'd prefer to purely operate on the capabilities API whenever some selection is made/changed, given we can pass everything to that endpoint anyway.

(We should probably still determine *which* carriers are available for a given destination here, and which options to pre-populate, but continue to drill down with continued calls to capabilities from there)

src/Context/Model/DynamicContext.php
```diff
@@ -70,6 +78,72 @@ public function __construct(?array $data = null)
...
+            'cc'     => $order->shippingAddress->cc ?? null,
+            'weight' => $this->computeTotalWeight($order),
+        ];
+
+        $projected = new CarrierCollection();
+
+        foreach ($carriers as $carrier) {
+            $capabilities = $validationService->getContextualCarrierCapabilities(
+                $carrier,
+                $shippingContext + ['packageTypes' => $carrier->packageTypes]
```
FreekVR: As noted elsewhere - avoid looping through carriers here. Just get a list of all available carriers (by essentially passing only destination to capabilities at this point) and their capabilities. What is available for each package type should be based on the selected package type (and other options) and then re-fetched when those change on the frontend.

src/Context/Model/DynamicContext.php
```diff
@@ -70,6 +78,72 @@ public function __construct(?array $data = null)
...
+
+        foreach ($order->lines ?? [] as $line) {
+            $lineWeight += (int) ($line->product->weight ?? 0) * max(1, (int) ($line->quantity ?? 1));
+        }
+
+        if ($lineWeight > 0) {
+            return $lineWeight;
+        }
+
+        return $order->physicalProperties->weight ?? null;
```
FreekVR: I feel this is also duplicated with earlier weight/package type related logic in an earlier PR.

src/App/Action/Backend/Context/FetchCapabilitiesAction.php
```diff
@@ -0,0 +1,84 @@
...
+     * @var \MyParcelNL\Pdk\Carrier\Repository\CarrierCapabilitiesRepository
+     */
+    private $repository;
+
+    /**
+     * @var \MyParcelNL\Pdk\Carrier\Service\CapabilitiesCarrierService
+     */
+    private $mapper;
+
+    public function __construct(CarrierCapabilitiesRepository $repository, CapabilitiesCarrierService $mapper)
```
FreekVR: `CapabilitiesCarrrierService` => `CapabilitiesToCarrierMapper` (Utils namespace)

src/Carrier/Service/CapabilitiesCarrierService.php
```diff
@@ -0,0 +1,148 @@
...
+            'options'          => $this->mapOptions($capability->getOptions()),
+        ];
+    }
+
+    /**
+     * @param  null|RefCapabilitiesResponseOptionsOptionsV2 $options
+     *
+     * @return array<string, array<string, mixed>>
+     */
+    private function mapOptions(?RefCapabilitiesResponseOptionsOptionsV2 $options): array
```
FreekVR: Can these mappers not be made simpler? Eg. by serializing to array and using known keys? This seems a bit overcomplicated and if this level of complexity is needed I'd much rather just return the `RefTypesCapabilities`-models directly from the API (whose shape is almost identical to the PDK Carrier)

src/Context/Model/DynamicContext.php
```diff
@@ -44,6 +48,16 @@ class DynamicContext extends Model
...
 
+    /**
+     * Contextual capabilities indexed by carrier name. Set when the context was
+     * built with an order; null otherwise. Used at serialization time to map each
+     * carrier's options into the PDK runtime shape (with requires/excludes,
+     * insurance amounts, etc.) via CapabilitiesCarrierService.
+     *
+     * @var array<string, RefCapabilitiesResponseCapabilityV2>|null
+     */
+    private $contextualCapabilities;
```
FreekVR: See my earlier comment, I don't like this pollution / overcomplication of the model here.

src/App/Action/Backend/Context/FetchCapabilitiesAction.php
```diff
@@ -0,0 +1,84 @@
...
+
+        if (! empty($body['packageType'])) {
+            // Body uses V2 format (e.g. "PACKAGE"); PACKAGE_TYPES_V2_MAP is keyed by lowercase
+            // legacy names mapped to V2 values. Look up by either side and fall back to the raw value.
+            $args['package_type'] = DeliveryOptions::PACKAGE_TYPES_V2_MAP[$body['packageType']]
+                ?? $body['packageType'];
+        }
+
+        if (isset($body['weight'])) {
+            $args['weight'] = (int) $body['weight'];
```
NabDevs: If I remember correctly, the generated SDK model doesn't have a top level `weight` field. Shouldn't this be  sent as `physical_properties.weight`?

src/Context/Model/DynamicContext.php
```diff
@@ -70,6 +84,95 @@ public function __construct(?array $data = null)
...
+        if (! $cc) {
+            return;
+        }
+
+        /** @var CapabilitiesValidationService $validationService */
+        $validationService = Pdk::get(CapabilitiesValidationService::class);
+
+        $capabilities = $validationService->indexByCarrier(
+            $validationService->getRepository()->getCapabilities([
+                'recipient' => ['cc' => $cc],
```
NabDevs: Should this initial admin context reuse the same package-type/order selection logic from the earlier capabilities PRs, or is it intentionally broad because the separate drill-down endpoint will narrow the carrier after JS selection changes?

FreekVR: refactored this branch to be a fix/extension of the capabilities proxy instead: JS-PDK uses order data + shipment preferences to calculate via the proxy what is available on the form

src/App/Action/Capabilities/CapabilitiesAction.php
```diff
@@ -41,36 +43,95 @@ public function handle(Request $request): Response
...
     /**
      * @param  \MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefCapabilitiesResponseCapabilityV2[] $results
+     * @param  bool                                                                                $filterOptions
+     *                                                                                                Apply the registered-options allowlist to each capability's options when true. Default false
+     *                                                                                                preserves the unfiltered SDK passthrough for existing callers.
      */
-    private function createSymfonyResponse(array $results): Response
+    private function createSymfonyResponse(array $results, bool $filterOptions = false): Response
     {
+        $body = json_decode(json_encode(['results' => $results]), true);
```
GravendeelJochem, line 122: We doen hier json_encode gevolgd door json_decode alleen om een array te krijgen om te filteren. Kan dit niet gewoon via array_map direct op $results?

FreekVR: Wat binnenkomt is een array met nested objecten/models. Dit is de makkelijkste manier om die allemaal "diep" te converten naar arrays. array_map zou alleen werken met een ingewikkeldere variant met recursions, dus denk ik niet persee beter

GravendeelJochem (approved):

NabDevs (approved):

PR #456, opened 2026-04-22
fix(INT-1261): combine Presta export compat and age-check toggle fixes

NabDevs:
## Summary
- keep `JsonResponse::create()` compatible with the Presta runtime
- apply carrier option cascades for age-check export
- fall back to global `*` carrier settings when a carrier-specific settings row is missing
- add regression coverage for Delivery Options settings fallback and age-check export


src/App/DeliveryOptions/Service/DeliveryOptionsService.php
```diff
@@ -204,6 +205,30 @@ private function createCarrierSettings(Carrier $carrier, PdkCart $cart, string $
...
+        if (null !== $storedSettings && $carrierSettings) {
+            return clone $carrierSettings;
+        }
+
+        $globalSettings = $allCarrierSettings->get(SettingsManager::KEY_ALL);
+
+        if ($globalSettings) {
+            return (clone $globalSettings)->fill(['id' => $carrier->carrier]);
+        }
+
```
FreekVR: I believe actually *using* the global settings is a new behavior, afaik upon installation all carriers have no defaults in presta currently?

PR #463, opened 2026-05-01
feat: handle shipment webhooks by order mode

NabDevs:
Adds order-mode aware handling to the PDK `ShipmentStatusChangeWebhook`.

- Add shared order mode constants to `AccountFeaturesServiceInterface`
- Use those constants in `PdkAccountFeaturesService`
- Handle shipment webhooks differently for:
  - Shipments only
  - Order v1
  - Order v2
- For Order v2, update existing local shipment data and order status without calling `UPDATE_SHIPMENTS`

INT-1331 and INT-1329


src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,389 @@
...
+     * @param  string[] $orderIds
+     * @param  array    $content
+     *
+     * @return void
+     */
+    private function updateShipmentsFromApi(array $orderIds, array $content): void
+    {
+        Actions::execute(PdkBackendActions::UPDATE_SHIPMENTS, [
+            'orderIds'                      => $orderIds,
+            'shipmentIds'                   => [$content['shipment_id']],
```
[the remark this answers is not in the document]
NabDevs: Updated this to resolve the shipment id through the existing helpers, including nested shipment payloads and alternative id keys. If no shipment id can be resolved, the webhook now logs and skips instead of reading shipment_id directly.

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    /**
+     * @param  array $content
+     *
+     * @return void
+     */
+    public function handleLabelCreated(array $content): void
+    {
+        if (AccountFeaturesServiceInterface::ORDER_MODE_V2 !== $this->accountFeaturesService->getOrderModeVersion()) {
```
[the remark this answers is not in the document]
NabDevs: Agreed. Normalized the order mode version to int here as well, matching handleStatusChange().

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,389 @@
...
+    {
+        $apiIdentifier = $this->getTrimmedValue($content, 'order_id');
+
+        if ('' === $apiIdentifier || ! method_exists($this->pdkOrderRepository, 'getByApiIdentifier')) {
+            return [];
+        }
+
+        /** @var null|\MyParcelNL\Pdk\App\Order\Model\PdkOrder $order */
+        $order = call_user_func([$this->pdkOrderRepository, 'getByApiIdentifier'], $apiIdentifier);
+
```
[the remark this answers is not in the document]
NabDevs: Added getByApiIdentifier() to PdkOrderRepositoryInterface and replaced the dynamic method check/call with a direct repository call.

GravendeelJochem (approved):

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,399 @@
...
+        $this->accountFeaturesService = $accountFeaturesService;
+        $this->pdkOrderRepository     = $pdkOrderRepository;
+    }
+
+    /**
+     * @param  array $content
+     *
+     * @return void
+     */
+    public function handleStatusChange(array $content): void
```
joerivanveen, line 42: This `$content` is merged sight unseen in different places with data already in the plugin. Should we not validate / transform this content to make sure it does not contain anything we can’t place / won’t use?

joerivanveen: Or do we have a reason to trust this content 100%...

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,399 @@
...
+
+        $this->updateShipmentsFromApi($orderIds, $content);
+    }
+
+    /**
+     * @param  array $content
+     *
+     * @return void
+     */
+    public function handleLabelCreated(array $content): void
```
joerivanveen, line 65: There is no longer an order status update (if set in settings) when the label is created it seems, as there was in the previous code.

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,399 @@
...
+        if ($requireExistingShipment && ! $existingShipment) {
+            $this->logSkippedWebhook('Skipping order v2 shipment status webhook for unknown shipment', $content);
+
+            return null;
+        }
+
+        $shipmentData = $this->getShipmentDataFromWebhook($shipmentContent, $order->externalIdentifier);
+
+        if ($existingShipment) {
+            $shipmentData = array_replace($existingShipment->toStorableArray(), $shipmentData);
```
joerivanveen: is this recursive? I’m afraid this may replace existing data with nothing when properties are not received in the webhook, sanity check.

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,399 @@
...
+
+        $existingShipment = $this->getShipmentFromOrder($order, $shipmentId);
+
+        if ($requireExistingShipment && ! $existingShipment) {
+            $this->logSkippedWebhook('Skipping order v2 shipment status webhook for unknown shipment', $content);
+
+            return null;
+        }
+
+        $shipmentData = $this->getShipmentDataFromWebhook($shipmentContent, $order->externalIdentifier);
```
joerivanveen: Is there a check whether the currently received shipment data is actually newer than the data we already have? E.g. when webhooks are received out of order for whatever reason?

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,399 @@
...
+    /**
+     * @param  array $content
+     *
+     * @return void
+     */
+    public function handleStatusChange(array $content): void
+    {
+        $orderModeVersion = (int) $this->accountFeaturesService->getOrderModeVersion();
+
+        if (AccountFeaturesServiceInterface::ORDER_MODE_V2 === $orderModeVersion) {
```
joerivanveen: Double check, it looks like we update / merge shipment data received in order V2, which is nice. But
1) do we display it somewhere?
2) the addBarcodeToNotes functionality is bypassed, which may be counterintuitive to a user when shipment data is updated elsewhere
Of am I missing something?

src/Account/Contract/AccountFeaturesServiceInterface.php
```diff
@@ -6,6 +6,12 @@
 
 interface AccountFeaturesServiceInterface
 {
+    public const ORDER_MODE_SHIPMENTS = 0;
```
joerivanveen, line 9: Maybe move the `FEATURE_*` constants from PdkAccountFeaturesService also to this interface for consistency?

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,399 @@
+<?php
```
joerivanveen, line 1: This file has a lot of private helpers, some probably already exist in the pdk or can be inlined. I leave that to you but for me it would improve the file to have less private helpers.

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,399 @@
...
+            : $content;
+    }
+
+    /**
+     * @param  array  $content
+     * @param  string $orderId
+     *
+     * @return array
+     */
+    private function getShipmentDataFromWebhook(array $content, string $orderId): array
```
joerivanveen: I agree with my AI buddy, which says:
Hydration belongs on the model, not in the service. getShipmentDataFromWebhook builds a snake-case→Shipment-shaped dictionary by hand. The PDK already supports this via Utils::changeArrayKeysCase + new Shipment([...]).    
  Pushing this into a Shipment::fromWebhookPayload(array $content): Shipment (or extending an existing transformer) would shrink the service substantially and remove the brittle manual key list.

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,399 @@
...
+use MyParcelNL\Pdk\App\Api\Backend\PdkBackendActions;
+use MyParcelNL\Pdk\App\Order\Contract\PdkOrderRepositoryInterface;
+use MyParcelNL\Pdk\App\Order\Model\PdkOrder;
+use MyParcelNL\Pdk\Facade\Actions;
+use MyParcelNL\Pdk\Facade\Logger;
+use MyParcelNL\Pdk\Settings\Model\OrderSettings;
+use MyParcelNL\Pdk\Shipment\Collection\ShipmentCollection;
+use MyParcelNL\Pdk\Shipment\Model\Shipment;
+
+final class ShipmentWebhookService
```
joerivanveen: Why is this final? Sanity check.

src/App/Webhook/Service/ShipmentWebhookService.php
joerivanveen: You might also look into moving the handling logic to three seperate files, one for each mode, and do mode detection here only. Maybe it will make the code clearer / cleaner?

joerivanveen (changes requested):
Nice! Some comments / questions. My AI buddy suggested these two improvements as well, with which I tend to agree:
  - Existing literal >= 2 for order mode — ExportOrderAction.php:110. Should use the new ORDER_MODE_V2 constant.                                                                                                                   
  - Test gaps — ShipmentStatusChangeWebhookTest.php/ShipmentLabelCreatedWebhookTest.php: no test pinning all three modes per case; no negative assertion that getByApiIdentifier is not called in v2; no skip-path coverage for v2 
  (unknown order, missing shipment id, requireExistingShipment branch, label-created with unknown order).

src/App/Webhook/Service/ShipmentWebhookService.php
NabDevs: Good point. I moved webhook payload hydration to Shipment::fromWebhookPayload(). It now only maps supported webhook fields.

src/App/Webhook/Service/ShipmentWebhookService.php
NabDevs: Good catch, this was missing. I changed the v2 shipment_label_created flow to use ShipmentUpdateService, so the existing order status update behavior is preserved.

src/App/Webhook/Service/ShipmentWebhookService.php
NabDevs: I changed this to use array_replace on the filtered shipment data. The webhook mapping is flat and null/empty values are filtered before merging, so missing fields should not overwrite existing shipment data.

src/App/Webhook/Service/ShipmentWebhookService.php
NabDevs: I made the behavior explicit in the tests: shipment_label_created preserves the existing shipment status, while shipment_status_change applies the incoming status. Without a timestamp/version in the payload, I think that is the safest distinction we can make here.

src/App/Webhook/Service/ShipmentWebhookService.php
NabDevs: Fixed, local v2 shipment updates now go through ShipmentUpdateService, so addBarcodeToNotes and order status updates are handled consistently with the existing shipment update flow.

src/Account/Contract/AccountFeaturesServiceInterface.php
NabDevs: Agreed. I moved the FEATURE_* constants to AccountFeaturesServiceInterface and kept aliases in PdkAccountFeaturesService for backwards compatibility.

src/App/Webhook/Service/ShipmentWebhookService.php
NabDevs: I moved shipment id/content parsing and webhook hydration out of ShipmentWebhookService and onto the Shipment model. I did not split the mode handling into separate classes for now, because after this cleanup the remaining branching is quite small.

src/App/Webhook/Service/ShipmentWebhookService.php
NabDevs: Agreed. This is now handled by Shipment::fromWebhookPayload() and Shipment::getIdFromWebhookPayload(). The mapping is explicit because the webhook supports multiple aliases for the same shipment fields.

src/App/Webhook/Service/ShipmentWebhookService.php
NabDevs: No specific reason. Is Removed.

src/App/Action/Backend/Shipment/UpdateShipmentsAction.php
```diff
@@ -58,75 +50,15 @@ public function handle(Request $request): Response
...
-
-        if ($orders->isNotEmpty()) {
-            $orders->updateShipments($shipments);
-            $this->pdkOrderRepository->updateMany($orders);
-
-            $this->addBarcodeNotes($shipments);
-
-            $this->updateOrderStatus($shipments, $request->get('orderStatus', OrderSettings::STATUS_ON_LABEL_CREATE));
-        }
+        $this->shipmentUpdateService->update(
```
joerivanveen, line 62: Nice.

src/Account/Service/PdkAccountFeaturesService.php
```diff
@@ -68,23 +41,23 @@ public function __construct(PdkAccountRepositoryInterface $accountRepository)
      */
     public function canUseOrderNotes(): bool
     {
-        return $this->hasFeature(self::FEATURE_ORDER_NOTES);
+        return $this->hasFeature(AccountFeaturesServiceInterface::FEATURE_ORDER_NOTES);
```
joerivanveen:
```suggestion
        return $this->hasFeature(self::FEATURE_ORDER_NOTES);
```

src/App/Action/Backend/Order/ExportOrderAction.php
```diff
@@ -107,7 +108,7 @@ protected function export(PdkOrderCollection $orders, Request $request): PdkOrde
             return $this->exportShipments($orders);
         }
 
-        if (AccountSettings::getOrderModeVersion() >= 2) {
+        if (AccountSettings::getOrderModeVersion() >= AccountFeaturesServiceInterface::ORDER_MODE_V2) {
```
joerivanveen, line 111: I would prefer to go through the modes we know (shipments, order v1, order v2) in order and throw otherwise. That way we know we need to do something here if V3 comes around (if ever). For consideration, not blocking imho.

src/Shipment/Model/Shipment.php
```diff
@@ -249,6 +302,125 @@ public function toStorableArray(): array
         return $array;
     }
 
+    /**
+     * @param  array  $payload
+     * @param  string $orderId
+     *
+     * @return array
+     */
+    private static function getWebhookPayloadData(array $payload, string $orderId): array
```
joerivanveen: I doubt these static helper functions should be part of the model, but not blocking as easily refactorable later.

joerivanveen (approved):

src/Account/Contract/AccountFeaturesServiceInterface.php
```diff
@@ -6,6 +6,22 @@
...
+
+    public const FEATURE_ORDER_MANAGEMENT = 'ORDER_MANAGEMENT';
+
+    public const FEATURE_LEGACY_ORDER_MANAGEMENT = 'LEGACY_ORDER_MANAGEMENT';
+
+    public const ORDER_MODE_SHIPMENTS = 0;
+
+    public const ORDER_MODE_V1        = 1;
+
+    public const ORDER_MODE_V2        = 2;
```
FreekVR: Aren't these mostly defined on the SDK side? Lets ensure we only define constants on the class that own the data and don't duplicate things outside of it. That would leave only the ORDER_MODE definitions here - those are the only ones we define ourselves.

see https://github.com/myparcelnl/sdk/blob/v11/src/Client/Generated/IamApi/Model/Feature.php

src/App/Order/Contract/PdkOrderRepositoryInterface.php
```diff
@@ -24,6 +24,11 @@ public function get($input): PdkOrder;
      */
     public function find($id): ?PdkOrder;
 
+    /**
+     * Find an order by its Order v1 API identifier.
+     */
+    public function getByApiIdentifier(string $uuid): ?PdkOrder;
```
FreekVR: might be a breaking chance that requires changes to wooc and presta. Did you check?

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,280 @@
...
+     * @param  array $content
+     *
+     * @return void
+     */
+    public function handleStatusChange(array $content): void
+    {
+        $orderModeVersion = (int) $this->accountFeaturesService->getOrderModeVersion();
+
+        if (AccountFeaturesServiceInterface::ORDER_MODE_V2 === $orderModeVersion) {
+            $this->updateOrderV2Shipment($content, true, $this->getStatusSettingFromWebhook($content));
```
FreekVR: getting the status directly from the webhook content is insecure without verifying the header containing the API key and comparing it to the stored API key (still not ideal, but this current implementation allows anyone to mutate the status, there's no current auth implementation here)

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,280 @@
...
+     */
+    public function handleLabelCreated(array $content): void
+    {
+        $orderModeVersion = (int) $this->accountFeaturesService->getOrderModeVersion();
+
+        if (AccountFeaturesServiceInterface::ORDER_MODE_V2 !== $orderModeVersion) {
+            return;
+        }
+
+        $this->updateOrderV2Shipment($content, false, OrderSettings::STATUS_ON_LABEL_CREATE, true);
```
FreekVR: this one also needs authentication verification - the other (v1/shipment) always result in an outgoing API call to the Core API, essentially confirming the data through that handshake. Here, we are using incoming data correctly to make assumptions, so we need to verify the webhook was called by our backoffice.

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        if (! $order->externalIdentifier) {
+            $this->logSkippedWebhook(
+                'Skipping order v2 shipment webhook for an order without external identifier',
+                $content
+            );
+
+            return null;
+        }
```
FreekVR: This is not really relevant for v2 orders - the e-commerce platform doesnt need to know the external ID, just the local ID, given it never does any outgoing requests and the incoming request will contain the local ID via shipment_reference_identifier. This may incorrectly reject valid order updates.

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,280 @@
...
+
+        $shipment = Shipment::fromWebhookPayload(
+            $content,
+            $order->externalIdentifier,
+            $existingShipment,
+            $preserveExistingStatus
+        );
```
FreekVR: As an alternative to this payload interpretation, we *can* still call the shipment API (shipment API will still work even in order v2 mode) that would workaround the issue with the payload not being verified / authenticated now and we can just use the existing logic (update triggers a new trusted call to fetch updated data from the shipment API) <-- I think this is the best fix here

The only "real" difference we need re. current situation:
- The differences in order/shipment/ID matching
- Ensure order v1 API is not called for order v2 mode

FreekVR (changes requested):
TL;DR: I think we should avoid creating/updating shipments from payload. We can also call "updateShipmentFromApi" for Order v2 to bypass the fact we don't verify incoming webhook with auth or payload verification.

Just parse the IDs as you currently do when v2 is active and then update from API rather than payload should work fine.

src/App/Webhook/Service/ShipmentWebhookService.php
joerivanveen: This would be my preference also. (Discussed this previously but thought it was not the way forward at this time.)

src/Account/Contract/AccountFeaturesServiceInterface.php
NabDevs: Changed this. The service now uses the SDK Feature constants where available, and the interface only contains the ORDER_MODE_* constants.

src/App/Order/Contract/PdkOrderRepositoryInterface.php
NabDevs: Agreed, removed it from the interface. The webhook service now only uses getByApiIdentifier defensively for order v1 when the repository supports it. Order v2 does not call this.

src/App/Webhook/Service/ShipmentWebhookService.php
NabDevs: The webhook no longer maps order status from the webhook payload. It only resolves the order/shipment identifiers and calls UPDATE_SHIPMENTS, so the status mapping is based on the shipment fetched from the API.

src/App/Webhook/Service/ShipmentWebhookService.php
NabDevs: Changed this as well. For order v2 label_created we now also trigger UPDATE_SHIPMENTS instead of using the webhook payload as shipment data.

src/App/Webhook/Service/ShipmentWebhookService.php
NabDevs: This check is gone now because the local payload merge path was removed. For order v2 we only use shipment_reference_identifier as the plugin order id and then fetch shipment data from the API.

src/App/Webhook/Service/ShipmentWebhookService.php
NabDevs: Done. The service now only handles order mode specific ID resolution and then delegates to UPDATE_SHIPMENTS.

src/App/Shipment/Service/ShipmentUpdateService.php
```diff
@@ -0,0 +1,134 @@
...
+            ->each(function (Shipment $shipment) use ($status) {
+                $resolvedStatus = $status ?? OrderSettings::getStatus((int) $shipment->status);
+
+                Logger::debug('Update status', [
+                    'orderId' => $shipment->orderId,
+                    'status'  => $resolvedStatus,
+                ]);
+                Actions::execute(PdkBackendActions::UPDATE_ORDER_STATUS, [
+                    'orderIds' => [$shipment->orderId],
+                    'setting'  => $resolvedStatus,
```
FreekVR, line 130: quick double check, "setting" is the correct key for passing a status to the action?

src/App/Webhook/Service/ShipmentWebhookService.php
```diff
@@ -0,0 +1,218 @@
...
+     * @param  array $content
+     *
+     * @return void
+     */
+    public function handleStatusChange(array $content): void
+    {
+        $orderModeVersion = (int) $this->accountFeaturesService->getOrderModeVersion();
+
+        $orderIds = $this->getOrderIds($content, $orderModeVersion);
+
```
FreekVR, line 47: nit: $orderMode service is only passed to `getOrderIds` and never used in this method. Could be removed as an arg and fetched from the accountFeatureService in `getOrderIds` itself.

FreekVR (approved):

PR #472, opened 2026-05-19
feat(account): default carrier follows shop's shipping rules

FreekVR:
## Summary

The default carrier used for shipments, returns, and the checkout delivery options now reflects the shipping rule configured for your shop in the MyParcel backoffice. Changes you make there propagate to the plugin on the next API-key save or via any flow that triggers an account refresh (debug environment switch, shop webhook, etc.).

Previously the default was a static value baked into the plugin's proposition configuration; updating it required a plugin release.

No new settings. Existing carrier preferences set per shipment are unaffected — the default only applies when no carrier is selected.

## Stack

This PR is stacked on top of [#471 (INT-1586)](https://github.com/myparcelnl/pdk/pull/471). It targets that branch directly so the diff shows only PR 2's changes. After #471 merges, I'll retarget this PR to `v4-capabilities`.

## Test plan

- [x] `docker compose run --rm php composer test:unit` — 2597 pass, 0 fail (PHP 7.4, 8.0, 8.4)
- [x] `docker compose run --rm php composer analyse` — PHPStan clean
- [x] After #471 merges + SDK release ships: smoke test against a real API key — save the key, verify `Pdk::get(AccountSettingsService::class)->getShop()->defaultCarrier` returns the expected V2 carrier name. Re-trigger via the debug switch action — value refreshes.
- [ ] Confirm no regression in checkout delivery options for existing orders that already have a carrier set

## What changed

- `Shop` gains `defaultCarrier` (V2 string, persisted) + virtual `defaultCarrierModel` (`?Carrier`).
- `UpdateAccountAction::setShopDefaultCarrier` fetches via PR 1's `ImplicationsService` and writes to Shop. On Service null (transient error or missing default), the previously persisted value is carried forward — flaky calls cannot wipe the cached default.
- All five historical callers of `PropositionService::getDefaultCarrier()` now read from `Shop`. Sites that expect a non-null `Carrier` (HasCarrierAttribute, DeliveryOptions) throw `RuntimeException` when no default exists; null-tolerant sites propagate null.
- `PropositionService::getDefaultCarrier()` is deleted, along with the now-dead `inbound`/`outbound` defaults from `PropositionContracts` and all three `proposition-*.json` configs.

## Behavioural changes worth flagging

- **Return-shipment fallback**: `PostReturnShipmentsRequest::ensureReturnCapabilities` previously _silently swapped_ to the default carrier (with a notification) when the user's carrier didn't support returns. With no default available, it now keeps the user's carrier — the export attempt will surface a more accurate error instead of a misleading notification.
- **`deliveryOptions.carrier` in context payloads**: `DeliveryOptions::toArray()` continues to emit the full resolved `Carrier` model (capabilities + options) — JS-PDK admin/checkout depend on that shape, so no shape change here. What changed is the resolution source: the getter now falls back to the shop's persisted default carrier, and the no-stored-carrier/no-default edge case emits `carrier: null` (or omits the key entirely when `Arrayable::SKIP_NULL` is set, e.g. `toArrayWithoutNull` / `toStorableArray` / `ENCODED`). `toStorableArray()` still returns the raw carrier name string as before. **JS-PDK reviewers: the only user-visible difference is that orders without an explicit carrier no longer get auto-filled with PostNL when no shop default exists — please confirm checkout/admin tolerates that.**

## Implementation note

`SettingsFactory::createDefault()` now calls `CarrierRepositoryInterface::all()` explicitly. The old code reached this via `PropositionService::getDefaultCarrier()` as a hidden side effect (the call walked through `CarrierRepository::findByLegacyId(1)` → `find('POSTNL')` → `all()` and primed the storage cache). Downstream tests that replace the bootstrap account with a slimmer shop depend on the cached `carrier:all` entry surviving. Surfacing the dependency explicitly avoids the same trap when the indirect path is removed.

Resolves INT-1479
Resolves INT-1504


src/App/Action/Backend/Account/UpdateAccountAction.php
```diff
@@ -85,6 +94,36 @@ public function handle(Request $request): Response
...
+        }
+
+        // Service returned null (transient error). Preserve the previously stored value if any.
+        $previousAccount = $this->pdkAccountRepository->getAccount();
+        $previousShop    = $previousAccount ? $previousAccount->shops->first() : null;
+
+        if ($previousShop && $previousShop->defaultCarrier) {
+            $shop->defaultCarrier = $previousShop->defaultCarrier;
+        }
+    }
```
FreekVR: Could be simplified, by swapping order.

If default carrier is null, just return (everything stays unchanged)
Any other value -> set the defaultCarrier on the shop

src/Context/Model/GlobalContext.php
```diff
@@ -77,15 +78,17 @@ public function __construct(?array $data = null)
...
-                'backofficeUrl'    => $proposition->applications['backoffice']['url'] ?? null,
-                'supportUrl'       => $proposition->applications['developerPortal']['url'] ?? null,
-                'localCountry'     => $proposition->countryCode,
-                'defaultCarrier'   => $defaultCarrier ? $defaultCarrier->carrier : null, // CONSTANT_CASE name
+                'name'           => $proposition->proposition->key,
+                'human'          => $proposition->proposition->name,
+                'backofficeUrl'  => $proposition->applications['backoffice']['url'] ?? null,
+                'supportUrl'     => $proposition->applications['developerPortal']['url'] ?? null,
+                'localCountry'   => $proposition->countryCode,
+                'defaultCarrier' => $defaultCarrier ? $defaultCarrier->carrier : null, // CONSTANT_CASE name
```
FreekVR: Check for usage(s), also in plugins/js. If there is none, we can and should remove this property given its not really part of the proposition config any more. 

Otherwise, we should expose it under a new attribute in the context ($this->attributes['shop']['defaultCarrier']) and encourage consumers to migrate to that attribute and deprecate the `proposition[defaultCarrier]`

src/Shipment/Model/DeliveryOptions.php
```diff
@@ -327,11 +328,21 @@ public function getCarrierAttribute(): Carrier
...
+
+            if ($default === null) {
+                throw new RuntimeException(
+                    $shop && $shop->defaultCarrier
+                        ? sprintf('Default carrier "%s" is not available in the repository', $shop->defaultCarrier)
+                        : 'No default carrier available; shop has no default carrier set'
+                );
+            }
+
+            return $default;
```
FreekVR: This code block seems to be repeated (not DRY). Perhaps a helper in a Shop- CarrierRepository would be useful?

tests/__snapshots__/CreateReturnShipmentsTest__it_creates_return_shipment_with_data_set_simple_domestic_shipment__1.json
```diff
@@ -162,7 +105,8 @@
                 "labelDescription": "standaard kenmerk"
             },
             "deliveryType": "standard",
-            "packageType": "package"
+            "packageType": "package",
+            "carrier": "POSTNL"
```
FreekVR: serialization of the carrier within DeliveryOptions seems to contain a breaking change. Consumers may rey on the whole carrier object being present. Suggest restoring it.

joerivanveen (approved):

GravendeelJochem (approved):
looks good to me

src/Shipment/Model/DeliveryOptions.php
```diff
@@ -381,6 +381,22 @@ public function isPickup(): bool
...
+        // default exists; detect that case upfront and emit the raw null attribute instead of
+        // throwing during serialization.
+        $hasStoredCarrier = (bool) $this->attributes[self::CARRIER];
+        $shop             = AccountSettings::getShop();
+        $hasShopDefault   = $shop && $shop->defaultCarrierModel;
+
+        if (! $hasStoredCarrier && ! $hasShopDefault) {
+            $array                = $this->except(self::CARRIER, $flags);
+            $array[self::CARRIER] = null;
+
```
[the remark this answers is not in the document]
FreekVR: Good catch. Fixed in 1cae6b6d — the early-return branch now skips the carrier=null assignment when SKIP_NULL is set, so toArrayWithoutNull / toStorableArray / ENCODED don't leak it. Added a regression test.

src/Shipment/Model/DeliveryOptions.php
```diff
@@ -381,6 +381,22 @@ public function isPickup(): bool
...
+            return Utils::filterNull([self::DATE => $this->getDateAsString()]) + $array;
+        }
+
         return Utils::filterNull([self::DATE => $this->getDateAsString()]) + parent::toArray($flags);
```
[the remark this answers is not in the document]
FreekVR: You're right, the description was off. toArray() still emits the full resolved Carrier model on purpose — JS-PDK consumers depend on that shape — and only toStorableArray() returns the raw string. Updated the PR description to reflect that.

tests/Unit/Shipment/Request/PostReturnShipmentsRequestTest.php
```diff
@@ -5,17 +5,26 @@
...
 use MyParcelNL\Pdk\Shipment\Model\Shipment;
 use MyParcelNL\Pdk\Tests\Uses\UsesMockPdkInstance;
 use MyParcelNL\Pdk\Tests\Uses\UsesNotificationsMock;
 use RuntimeException;
 use function MyParcelNL\Pdk\Tests\factory;
+use function MyParcelNL\Pdk\Tests\mockPdkProperties;
 use function MyParcelNL\Pdk\Tests\usesShared;
 use MyParcelNL\Pdk\Tests\Uses\UsesAccountMock;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefCapabilitiesSharedCarrierV2;
+use MyParcelNL\Sdk\Client\Generated\CoreApi\Model\RefTypesCarrier;
```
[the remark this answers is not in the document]
FreekVR: Removed in 1cae6b6d, thanks.

PR #476, opened 2026-05-27
fix: include all enabled carriers in carrierSettings 

GravendeelJochem:
INT-1599

### DIFF
--- a/tests/Unit/App/DeliveryOptions/Service/DeliveryOptionsServiceCapabilitiesTest.php
+++ b/tests/Unit/App/DeliveryOptions/Service/DeliveryOptionsServiceCapabilitiesTest.php
@@ -133,7 +133,7 @@ function enqueueCapabilitiesPerType(array $responsesPerType): void
     }
 }
 
-it('excludes carrier when package type is not in capabilities', function () {
+it('includes enabled carrier in carrierSettings even when absent from capabilities response', function () {
     storeCarrierSettings([
         RefCapabilitiesSharedCarrierV2::POSTNL      => true,
         RefCapabilitiesSharedCarrierV2::DHL_FOR_YOU  => true,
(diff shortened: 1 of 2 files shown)


tests/Unit/App/DeliveryOptions/Service/DeliveryOptionsServiceCapabilitiesTest.php
```diff
@@ -167,9 +168,14 @@ function enqueueCapabilitiesPerType(array $responsesPerType): void
...
 
+    // Both carriers must appear — the widget's own capabilities proxy call determines
+    // actual per-destination availability; carrierSettings only carries merchant config.
     expect($result['packageType'])->toBe(DeliveryOptions::PACKAGE_TYPE_PACKAGE_NAME)
         ->and($result['carrierSettings'])->toHaveKey($postnlId)
-        ->and($result['carrierSettings'])->not->toHaveKey($dhlId);
+        ->and($result['carrierSettings'])->toHaveKey($dhlId)
+        // PostNL was in capabilities → contractId propagated; DHL_FOR_YOU was not → null.
+        ->and($result['carrierSettings'][$postnlId]['contractId'])->toBe(100)
+        ->and($result['carrierSettings'][$dhlId]['contractId'])->toBeNull();
```
FreekVR, line 178: This seems more like a workaround for an incorrect capabilities result in the plugin rather than an actual fix.

The current behavior is important for prestashop. We show a list of possible carriers *before* the DO JS widget does the API call to check the possibilities for the carrier.

In prestashop, this would result in the user being able to click a carrier -> do opens -> says no results. The carrier should be filtered from the list of possible carriers before sending it off to the DO - but it seems *something* is causing that list of carriers to be incomplete or for an incorrect package type?

FreekVR (changes requested):

PR #491, opened 2026-06-10
fix: don't crash on incomplete carrier data, just filter it out silently

FreekVR:


joerivanveen (approved):

tests/Unit/Account/Service/AccountSettingsServiceTest.php
```diff
@@ -109,6 +109,38 @@
...
+            'shops'      => [
+                [
+                    'id'         => '666',
+                    'accountId'  => '1112223',
+                    'platformId' => 1,
+                    'name'       => 'PartialShop',
+                    'carriers'   => [
+                        ['carrier' => RefCapabilitiesSharedCarrierV2::POSTNL],
+                        // Partial/legacy data with no carrier name: $carrier->carrier resolves to null.
+                        ['packageTypes' => ['PACKAGE']],
```
NabDevs, line 130: Shouldn't this also test a valid 'carrier' key, without a supported carrier ref/value? e.g.: `['carrier' => ['blabla']` ?

NabDevs: Never mind, I found that exact test on line 77 :)

NabDevs (approved):

PR #497, opened 2026-07-06
feat: add SPRING and VIA_TIM carriers (INT-1683)

GravendeelJochem:
INT-1683

## Summary

Adds the two new carriers **SPRING** (id 19) and **VIA_TIM** (id 20), introduced in `myparcelnl/sdk` `11.0.0-beta.28`.

- Bump `myparcelnl/sdk` to `^11.0.0-beta.28@beta`.
- Register both carriers in `Carrier::CARRIER_NAME_TO_LEGACY_MAP` and `Carrier::CARRIER_NAME_ID_MAP` (+ legacy-name constants), so they pass `Carrier::isSupported()` and surface in carrier settings. This follows the [wiki](https://myparcelnl.atlassian.net/wiki/spaces/IG/pages/495124481) §1.1–1.2 (SDK first, then the PDK maps).

## SDK bump fallout (fixed in this PR)

The beta.28 regeneration also required two unrelated-to-carriers fixes to keep the bump clean:

1. **`postCapabilities` / `postCapabilitiesContractDefinitions` argument order** — the generated client now takes the request body first and the user agent second. The previous call order sent the user agent as the request body. Both call sites in `CapabilitiesService` fixed; the rest of the `SdkApi` layer was audited and these were the only affected sites.
2. **`cooled_delivery` removed from the capabilities V2 options model** (but kept on `RefShipmentShipmentOptions`). `CooledDeliveryDefinition` now returns `null` for its capabilities key (like `ExcludeParcelLockers`) instead of reading a removed `attributeMap` key. It remains a settable shipment option but is no longer capability-gated. Test assertions updated accordingly.

Snapshots regenerated to reflect the new carriers and the removed `cooled_delivery` capability.

## Not in this PR (needed for end-to-end, per wiki §1.3–1.4)

- Plugins (WooCommerce/PrestaShop): bump their `myparcelnl/pdk` dependency to the release containing this change.
- Delivery Options: add `SPRING`/`VIA_TIM` to the `CarrierName` enum in `@myparcel-dev/constants` + sandbox.
- Translations (`carrier_spring`, `carrier_via_tim`): Google Sheet rows, not code.

## Test plan

- [x] `yarn run test:unit` — 2642 passed, 0 failed, 19 skipped
- [x] integration tests green


tests/__snapshots__/SettingsViewTest__it_gets_settings_view_with_data_set_carrier_settings__1.json
```diff
@@ -23435,10 +23357,3568 @@
...
+                    "$component": "CurrencyInput",
+                    "label": "settings_carrier_price_delivery_type_pickup",
+                    "description": "settings_carrier_price_delivery_type_pickup_description"
+                }
+            ]
+        },
+        {
+            "id": "VIA_TIM",
+            "title": "carrier_via_tim",
+            "description": "settings_view_carrier_via_tim_description",
```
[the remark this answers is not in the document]
GravendeelJochem: Good catch — the intended key is `carrier_via_tim` (with underscore), which is what the PDK actually emits: the translation key is derived from the V2 carrier name via `Str::snake(Str::lower($carrier))` in `CarrierSettingsItemView`, so `VIA_TIM` → `carrier_via_tim` and `settings_view_carrier_via_tim_description`. `SPRING` → `carrier_spring`.

The `carrier_viatim` in the PR description was a typo in the description text; the code/snapshots are correct. Downstream translation work (Google Sheet rows / `@myparcel-dev/constants`) should use `carrier_via_tim` and `carrier_spring`. Updated the PR description to match.

src/App/Options/Definition/CooledDeliveryDefinition.php
```diff
@@ -15,9 +14,14 @@ public function getShipmentOptionsKey(): ?string
...
 
+    /**
+     * Cooled delivery was removed from the capabilities V2 options as of SDK 11.0.0-beta.28;
+     * it remains a settable shipment option ({@see RefShipmentShipmentOptions}) but is no
+     * longer capability-gated, so it has no capabilities key (like ExcludeParcelLockers).
+     */
     public function getCapabilitiesOptionsKey(): ?string
     {
-        return RefCapabilitiesContractDefinitionsResponseOptionsOptionsV2::attributeMap()['cooled_delivery'];
+        return null;
```
FreekVR: We need to have a better look at this. If its not in capabilities, it will never bubble down as an available option and quietly be unavailable for use.

Was it perhaps renamed or moved to another class? Why was it removed?

Either we completely remove the option if its not in capabilities, or we fix it if its available under some other name/definition.

FreekVR (changes requested):

src/App/Options/Definition/CooledDeliveryDefinition.php
GravendeelJochem: Checked the SDK tags: `cooled_delivery` was in the capabilities options model up to beta.23 and removed in **beta.24**  not renamed or moved, and absent from every capabilities options model in beta.28. So there is no other name/definition to point to.

Went with fully removing the option (definition + registration + model property + tests/snapshots). Pushed in 08cec2fe.

joerivanveen (approved):

FreekVR (approved):
I am still double checking whether cooled delivery should actually have been removed from capabilities. But since the current state is that its not available, I'll approve in this form.

PR #498, opened 2026-07-06
fix: move fetch context endpoint to backend

joerivanveen:
INT-1692


FreekVR (approved):

config/actions.php
```diff
@@ -105,6 +97,11 @@
     ],
 
     PdkEndpoint::CONTEXT_BACKEND => [
+        PdkSharedActions::FETCH_CONTEXT                 => [
```
FreekVR: We should move the FETCH_CONTEXT out of the PdkSharedActions list and to PdkBackendActions

GravendeelJochem (approved):

config/actions.php
joerivanveen: Done

PR #499, opened 2026-07-08
fix: harden container cache handling

NabDevs:
This addresses failures seen after the WooCommerce 6.6.0 release where stale container/cache state or symlinked webroots with `open_basedir` restrictions could cause fatal errors.

- Resolve PDK root/config paths via `realpath()` with fallbacks
- Prevent `pdkVersion` lookup from failing the application when `composer.json` cannot be read
- Key compiled container/APCu definition caches by app version
- Clear all compiled container cache files recursively, including OPcache invalidation

Trade-offs

- Old APCu cache entries are not actively removed. They expire/evict by themselves.
- On upgrade, the container may be compiled one extra time. I kept this simple for the hotfix.
- The cache key now uses the app version. If needed, we can also add a path hash to make it unique per install.
- I did not change the compiled container class name. That was already shared before this PR and feels out of scope here.

INT-1709

### DIFF
--- a/config/pdk-dependencies.php
+++ b/config/pdk-dependencies.php
@@ -32,6 +32,10 @@ return [
     'pdkNextMajorVersion'       => factory(function (): string {
         $version = Pdk::get('pdkVersion');
 
+        if (! preg_match('/^\d+\./', $version)) {
+            return 'unknown';
+        }
+
         return (int) explode('.', $version)[0] + 1 . '.0.0';
     }),
 
--- a/config/pdk-dependencies.php
+++ b/config/pdk-dependencies.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 
 use MyParcelNL\Pdk\Base\FileSystemInterface;
+use MyParcelNL\Pdk\Facade\Logger;
 use MyParcelNL\Pdk\Facade\Pdk;
 use MyParcelNL\Pdk\Frontend\Contract\ScriptServiceInterface;
 
@@ -22,6 +23,14 @@ return [
 
             return $composerJson['version'] ?? 'unknown';
         } catch (Throwable $e) {
+            try {
+                Logger::warning('Unable to determine PDK version', [
+                    'error' => $e->getMessage(),
+                ]);
+            } catch (Throwable $loggerException) {
+                // The version fallback must never break boot, even if logging is unavailable.
+            }
+
             return 'unknown';
         }
     }),
--- a/config/pdk-dependencies.php
+++ b/config/pdk-dependencies.php
@@ -11,13 +11,19 @@ use function DI\value;
 
 return [
     /**
-     * The current version of the pdk according to the composer.json file.
+     * The current version of the pdk according to the composer.json file. Falls back to "unknown"
+     * if the file cannot be read; this value is informational (user agent headers) and reading it
+     * must never take down the application. See myparcelnl/woocommerce#1664.
      */
     'pdkVersion'                => factory(function (FileSystemInterface $fileSystem): string {
-        $rootDir      = Pdk::get('rootDir');
-        $composerJson = json_decode($fileSystem->get("$rootDir/composer.json"), true);
+        try {
+            $rootDir      = rtrim(Pdk::get('rootDir'), '/');
+            $composerJson = json_decode($fileSystem->get("$rootDir/composer.json"), true);
 
-        return $composerJson['version'];
+            return $composerJson['version'] ?? 'unknown';
+        } catch (Throwable $e) {
+            return 'unknown';
+        }
     }),
 
     /**
--- a/src/Base/Factory/PdkFactory.php
+++ b/src/Base/Factory/PdkFactory.php
@@ -22,6 +22,11 @@ class PdkFactory implements PdkFactoryInterface
     /**
      * @var null|string
      */
+    protected static $cacheVersion;
+
+    /**
+     * @var null|string
+     */
     protected static $mode;
 
     public function __construct() { }
@@ -42,6 +47,20 @@ class PdkFactory implements PdkFactoryInterface
         Facade::setPdkInstance($pdk);
 
         return $pdk;
+    }
+
+    /**
+     * Set the app version to key the container caches by. Without this, caches persist across
+     * up- and downgrades (most notably the APCu definition cache, which survives replacing the
+     * plugin directory) and poison the compiled container with definitions from another version.
+     *
+     * @param  null|string $version
+     *
+     * @return void
+     */
+    public static function setCacheVersion(?string $version): void
+    {
+        self::$cacheVersion = null === $version ? null : preg_replace('/[^\w.-]+/', '_', $version);
     }
 
     /**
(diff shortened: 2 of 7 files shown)


config/pdk-dependencies.php
```diff
@@ -11,13 +11,19 @@
...
-        $composerJson = json_decode($fileSystem->get("$rootDir/composer.json"), true);
-
-        return $composerJson['version'];
+        try {
+            $rootDir      = rtrim(Pdk::get('rootDir'), '/');
+            $composerJson = json_decode($fileSystem->get("$rootDir/composer.json"), true);
+
+            return $composerJson['version'] ?? 'unknown';
+        } catch (Throwable $e) {
+            return 'unknown';
```
FreekVR, line 34: I'd suggest logging a warning eg. "Unable to determine plugin version, {e->getMessage()}" so this doesn't fail quietly (hard to debug down the line I think)

src/Base/Factory/PdkFactory.php
```diff
@@ -19,6 +19,11 @@ class PdkFactory implements PdkFactoryInterface
     private const MODES       = [Pdk::MODE_PRODUCTION, Pdk::MODE_DEVELOPMENT];
     private const CONFIG_PATH = __DIR__ . '/../../../config';
```
FreekVR, line 20: Double check, shouldn't the `realpath()` be applied here, too?

src/Base/Factory/PdkFactory.php
```diff
@@ -44,6 +49,20 @@ public static function create(...$config): PdkInterface
...
+     * plugin directory) and poison the compiled container with definitions from another version.
+     *
+     * @param  null|string $version
+     *
+     * @return void
+     */
+    public static function setCacheVersion(?string $version): void
+    {
+        self::$cacheVersion = null === $version ? null : preg_replace('/[^\w.-]+/', '_', $version);
+    }
```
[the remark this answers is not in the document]
NabDevs: It normally comes from the plugin version, but since we use it as a directory segment I hardened it anyway. Unsafe values now fall back to `default`.

config/pdk-dependencies.php
```diff
@@ -11,13 +11,19 @@
 
 return [
     /**
-     * The current version of the pdk according to the composer.json file.
+     * The current version of the pdk according to the composer.json file. Falls back to "unknown"
+     * if the file cannot be read; this value is informational (user agent headers) and reading it
+     * must never take down the application. See myparcelnl/woocommerce#1664.
      */
```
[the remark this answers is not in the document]
NabDevs: Agreed, fixed. If `pdkVersion` is `unknown`, `pdkNextMajorVersion` now returns `unknown` as well.

src/Base/Factory/PdkFactory.php
```diff
@@ -44,6 +49,29 @@ public static function create(...$config): PdkInterface
...
+    /**
+     * Set the app version to key the container caches by. Without this, caches persist across
+     * up- and downgrades (most notably the APCu definition cache, which survives replacing the
+     * plugin directory) and poison the compiled container with definitions from another version.
+     *
+     * @param  null|string $version
+     *
+     * @return void
+     */
+    public static function setCacheVersion(?string $version): void
```
FreekVR, line 61: Looking at the architecture across the PRs, I have some concerns. It seems that several concepts / workarounds are being conflated:

- The [plugin duplicates ](https://github.com/myparcelnl/woocommerce/pull/1666/changes) part of the PDKs existing (now updated) cache clearning, because apparently the PDK cannot be reliably booted with a stale cache, so the PDKs existing cache clear would not run
- The PDK however retains its own version-gated cache clear via the `install()` method. This effective clears the cache *twice* on a plugin update (once before boot, once after)
- The *PDK*- (not *APP* version as implied in the comment) version prefix introduced here adds another layer which effectively also invalidates caches for previous/other versions of the PDK (not the plugin) which seems redundant if the cache clears are doing their job. However, we cannot use the prefix by itself, because then older versions of the cache would remain in place. I'd suggest removing the versioned cache prefix.

I'd argue the PDK should own the cache clears. There are several locations within the PdkBoostrapper and PdkFactory which should be able to clear caches before the full boot / DI setup occurs. The only thing we need to figure out is how to detect a plugin's version changed. Only the "current" (`composer.json`-based) plugin version is passed - and we don't have access to the full suite of utilities that `install()` uses to determine the previously present plugin version from the DB.

Note: Normally a deploy job would clear caches on any deploy before any HTTP request/php script runs. This is what PHP-DI recommends when using the CompiledContainer and APC caches. We need to ensure that the cache clear here immediately "works" on the first request, and doesn't require a page refresh/second request before it effectively takes effect. We don't really have any other options given we don't own the infra. This could be a rationale for using the prefix after all, but I'm still note sure if that prefix was needed.

An option to explore:
The cache can hold the information about which version it was generated for (and maybe already does!). Similar to the prefix idea, but retrievable and doesn't require any key prefixing. If feasable, seems the cleanest? We'd only use this data to determine whether to clear the cache, not as a validator that decides whether to fetch it to begin with (so substantially different from the prefix concept there)

```
/* NOTE: just an example, perhaps the app version is ALREADY stored in the container and we can read it without adding it. */


// Fetch the cached app version (which version is the cache built for?) *before* boot
$compiledVersion = $container->has('app.version') ? $container->get('app.version') : null;

[...]


// Inject the version definition so it gets baked into the compiled file (may already be the case!)
    $containerBuilder->addDefinitions([
        'app.version' => $appInfo->version,
    ]);
```

TL;DR: compare the provided (in AppInfo) and container/cache-stored versions before boot but within the PDK, and clear the caches if they differ. No prefix or plugin-side code needed.

src/Base/Pdk.php
FreekVR, line 21: Check: shouldn't this use realpath() too?

FreekVR (changes requested):

src/Base/Factory/PdkFactory.php
```diff
@@ -77,12 +105,23 @@ protected function getMode(): string
      */
     protected function setupCache(ContainerBuilder $builder): void
     {
+        $cacheVersion = $this->getCacheVersion();
+        $cacheDir     = Pdk::CACHE_DIR . '/' . $cacheVersion;
+
         if (SourceCache::isSupported()) {
-            $builder->enableDefinitionCache('pdk-definition-cache');
+            $builder->enableDefinitionCache("pdk-definition-cache-$cacheVersion");
```
FreekVR, line 112: Note: This versions the *cache* but NOT the compiled container.

config/pdk-dependencies.php
NabDevs: Agreed, added a warning when the PDK version cannot be determined.

src/Base/Factory/PdkFactory.php
NabDevs: I now resolve this through `realpath(self::CONFIG_PATH) ?: self::CONFIG_PATH` at the point where the config files are added. The constant itself cannot call `realpath()`.

src/Base/Pdk.php
NabDevs: Same here: `CACHE_DIR` stays a static constant, but cache reads/writes now go through `Pdk::getCacheDir()`, which resolves the path when possible and falls back when `.cache` does not exist yet.

src/Base/Factory/PdkFactory.php
NabDevs: Fixed. The compiled container class name now includes the sanitized cache version as well, so the cache dir and compiled class stay in sync.

src/Base/Factory/PdkFactory.php
NabDevs: The Woo side is now dropped. I checked your proposed solution, but chose to keep the version prefix because it avoids loading a potentially stale/broken compiled container or stale cached container data just to decide whether it should be cleared. So the prefix gives us isolation on the first request, and `clearCache()` still handles cleanup of old versions. If you see this differently, lmk.

NabDevs: Small clarification: this prefix is the app/plugin version from `PdkBootstrapper::boot()`, not the PDK version. So it changes when a customer installs a new plugin version.

The main reason I kept it is that a marker inside the compiled container does not really cover the case we saw: after a plugin update that file can be gone, while stale in-memory cache can still exist. Then the first fresh compile can still pick up stale data. The prefix avoids that by using a new cache key immediately.

FreekVR (approved):

GravendeelJochem (approved):

PR #504, opened 2026-07-10
feat: correct B2B/B2C capabilities via an address-based isBusiness flag

FreekVR:
Adds support for explicit `isBusiness` flags based on the adress (a filled company name denotes a B2B shipment, an empty one sets it to false and makes it B2C) and fixes issue where it was set to `true` instead of `null`

## Note

- Please **rebase-merge** so the fix and feat stay as separate commits.

## Follow-ups (separate PRs, other repos)

- **js-pdk** (admin): forward `order.shippingAddress.isBusiness` on the proxy capabilities call.
- **delivery-options**: config-provided `isBusiness` boolean on the recipient (no company field on the widget address).
- **PrestaShop**: map `company` into the cart array + reuse the rule in the carrier-list hook. **WooCommerce**: test-only (its adapter already supplies the company).

Fixes INT-1690


src/App/DeliveryOptions/Service/DeliveryOptionsService.php
```diff
@@ -231,7 +231,9 @@ private function getValidCarrierOptions(PdkCart $cart): array
     {
         $allCarriers           = $this->carrierRepository->all();
         $carrierSettings       = Settings::get(CarrierSettings::ID);
-        $cc                    = $cart->shippingMethod->shippingAddress->cc ?? null;
+        $shippingAddress       = $cart->shippingMethod->shippingAddress;
+        $cc                    = $shippingAddress->cc ?? null;
+        $isBusiness            = $shippingAddress->isBusiness;
```
GravendeelJochem, line 236: De ?? null op $shippingAddress->cc suggereert dat het adres null kan zijn, maar $shippingAddress->isBusiness heeft geen guard. Is dit bewust?

FreekVR: `isBusiness` cast naar een `bool` op het `Addres` model, met een default `false`-value. Daardoor wordt het niet gedropped door allerlei "withoutNull" guards die er wel voor kunnen zorgen dat strings met een default `null`-waarde (zoals `cc`) niet altijd als attribuut op het model zitten.

GravendeelJochem (approved):

src/Base/Model/Address.php
```diff
@@ -82,6 +85,35 @@ public function __construct(?array $data = null)
...
+
+        // INT-1690 support: settle the business flag deterministically after fill. A supplied
+        // company name is authoritative — it always dictates the flag, so a stale/round-tripped
+        // isBusiness cannot override it. An explicit isBusiness only applies when no company is
+        // given. `company` is not an Address attribute, so it is read here only to derive the flag
+        // and is never stored or serialised on a bare Address — it lives on ContactDetails, the
+        // PII-bearing model.
+        if (is_array($convertedData)) {
+            if (array_key_exists('company', $convertedData)) {
+                $this->attributes['isBusiness'] = self::deriveIsBusiness($convertedData['company']);
```
NabDevs, line 97: I think `company: null` overrides `isBusiness: true` here?  Since isBusiness: true without a company is a valid use case, shouldn’t it be preserved?

NabDevs (changes requested):
Looks good, 1 potential issue that needs to be checked I think.

src/Base/Model/Address.php
FreekVR: From a purely architectural standpoint that does make more sense, but I decided against it.

Most "B2B" shipments will require a mandatory company name in the final API calls to create the shipment. This is like a guard to not offer those carriers if that mandatory field is not available on the Order.

It's structurally not exactly the right place to do that, but we don't have a dedicated validation layer other than capabilities, and this way we let capabilities decide for us even though it doesn't actually do address / shipment input validation.

NabDevs: Aaah thanks, that makes sense. I was assuming isBusiness should always be preserved. Preventing B2B carriers from being offered when there's no company name seems like a good trade-off for now. Approved!

NabDevs (approved):

PR #506, opened 2026-07-15
fix(settings): always render settings page when account fetch fails

RichardPerdaan:
## Problem

When fetching the account from the API fails (API down, network error, unexpected 5xx), the exception bubbled up into `FrontendRenderService::renderTemplate()`, which catches all throwables and renders an **empty string**. As a result the entire plugin settings page disappeared — including the API key form the merchant needs to fix the problem. The init script suffered the same fate via `DynamicContext`, so the admin app would not even boot.

Reported from WooCommerce, but this affects every PDK-based plugin.

## Solution

- **`AbstractPdkAccountRepository::getAccount()`**: treat a failed account fetch as "no account" — log the error, push an error notification and return `null`, matching the existing invalid-api-key semantics. Since `DynamicContext` uses the same repository, this fixes both the settings page render and the init script in one place.
- **`PluginSettingsViewContext`**: defensively catch throwables while building the settings views, so a single failing view can never blank the page either.

The admin frontend (js-pdk) already handles `account: null` correctly: `PluginSettings.vue` renders `<AccountSettings />` unconditionally and opens the API key tab, hiding only the remaining settings forms. So with this change the API key block is **always** visible, with an error notification explaining what went wrong.

## Tests

- New: `tests/Unit/App/Account/Repository/AbstractPdkAccountRepositoryTest.php`
  - API error → `getAccount()` returns `null` and adds a notification (fails without this fix)
  - API error → `createContexts([ID_PLUGIN_SETTINGS_VIEW])` no longer throws
- Full suite: 2678 passed; pre-existing local failures in `ZipService`/`AddressesApiService`/`DownloadLogs` also fail on `main` (PHP version issue, unrelated)
- PHPStan: no errors on changed files

### DIFF
--- a/src/App/Account/Repository/AbstractPdkAccountRepository.php
+++ b/src/App/Account/Repository/AbstractPdkAccountRepository.php
@@ -8,8 +8,12 @@ use MyParcelNL\Pdk\Account\Model\Account;
 use MyParcelNL\Pdk\Account\Repository\AccountRepository;
 use MyParcelNL\Pdk\App\Account\Contract\PdkAccountRepositoryInterface;
 use MyParcelNL\Pdk\Base\Repository\Repository;
+use MyParcelNL\Pdk\Facade\Logger;
+use MyParcelNL\Pdk\Facade\Notifications;
+use MyParcelNL\Pdk\Notification\Model\Notification;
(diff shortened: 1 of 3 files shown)


RichardPerdaan:
Closing this PR.

PR #508, opened 2026-07-20
fix(settings): render same-day settings for carriers (trunkrs) with same-day delivery type

FreekVR:
renders the same-day delivery settings for carriers that expose same-day as a delivery type instead of a shipment option.

Some carriers (like Trunkrs) advertise same-day delivery as a delivery type in their contract, while others (like DHL For You) advertise it as a shipment option. The carrier settings view only checked the shipment option, so for delivery-type carriers the cutoff time field never rendered and the same-day cutoff silently stayed at the 10:00 default. When same-day is all the carrier offers, that hides the carrier from checkout after the cutoff with no way to fix it in the admin. The same-day section now renders for both representations and is the single owner of these fields. This also removes a duplicate same-day toggle that was rendered for shipment-option carriers.

Fixes INT-1561

### DIFF
--- a/src/Frontend/View/CarrierSettingsItemView.php
+++ b/src/Frontend/View/CarrierSettingsItemView.php
@@ -486,21 +486,33 @@ class CarrierSettingsItemView extends AbstractSettingsView
         // delivery type (e.g. Trunkrs), depending on the carrier's contract. This
         // section is the single owner of the same-day fields for both representations;
         // getDeliveryTypeSettings() and getShipmentOptionsSettings() skip same-day.
+        $definition = new SameDayDeliveryDefinition();
+
         $hasSameDayDeliveryType = $this->carrier->deliveryTypes
             && in_array(RefTypesDeliveryTypeV2::SAME_DAY, $this->carrier->deliveryTypes, true);
 
-        if (
-            ! $hasSameDayDeliveryType
-            && ! $this->carrierValidationService->supportsShipmentOption($this->carrier, SameDayDeliveryDefinition::class)
-        ) {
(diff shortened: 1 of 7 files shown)


src/Frontend/View/CarrierSettingsItemView.php
```diff
@@ -482,7 +482,17 @@ function (FormOperationBuilder $builder) {
      */
     private function getSameDayDeliverySettings(): array
     {
-        if (!$this->carrierValidationService->supportsShipmentOption($this->carrier, SameDayDeliveryDefinition::class)) {
+        // Same-day is exposed either as a shipment option (e.g. DHL For You) or as a
+        // delivery type (e.g. Trunkrs), depending on the carrier's contract. This
+        // section is the single owner of the same-day fields for both representations;
+        // getDeliveryTypeSettings() and getShipmentOptionsSettings() skip same-day.
+        $hasSameDayDeliveryType = $this->carrier->deliveryTypes
```
NabDevs, line 491: Looks good to me, would prefer below for readability though, but just a preference I think.  

```
$hasSameDayDeliveryType = $this->carrier->deliveryTypes...                                                                                                                                                                              
$hasSameDayShipmentOption = $this->carrierValidationService...
                                                                                                                                                                                                                          
if (!$hasSameDayDeliveryType && !$hasSameDayShipmentOption) {                                                                                                                                                                          
      return [];                                                                                                                                                                                                                         
  }     
```

NabDevs (approved):

GravendeelJochem (approved):

PR #510, opened 2026-07-24
feat(checkout): pass the cart's shipment options to the delivery options widget

FreekVR:
Adds the cart's calculated shipment options to the checkout context, and fixes the age check (18+) rules on the carrier settings screen.

The checkout context gets a new optional cartShipmentOptions key (a sibling of config): per carrier, the widget's camelCase option name → boolean, calculated by running the cart through the same pipeline an order export uses. The delivery options widget shows those options (like the ones 18+ requires) as checked and locked for the consumer; without the key the widget behaves as before. js-pdk forwards the key to the widget in myparcelnl/js-pdk#381.

On the carrier settings screen the 18+ rules now lock signature and only recipient only when 18+ is actually on — the lock previously also matched "inherit". Test fixtures now carry the real capability rules for age verification (requires signature and only recipient, excludes receipt code), with regression tests pinning the export path.

Merge order for INT-1596:
1. myparcelnl/delivery-options#480 — its release adds cartShipmentOptions to the widget input.
2. myparcelnl/js-pdk#381 — bump @myparcel-dev/delivery-options there afterwards to turn its typecheck green.
3. This PR — independent, can merge any time.

Fixes INT-1596
Subtask: INT-1732


src/Context/Model/CheckoutContext.php
```diff
@@ -56,10 +66,23 @@ public function __construct(?array $data = null)
...
-            'settings' => [
+        foreach ($deliveryOptionsService->createCartShipmentOptions($cart)->all() as $identifier => $shipmentOptions) {
+            $cartShipmentOptions[$identifier] = ShipmentOptions::toBooleanOptions($shipmentOptions);
+        }
+
+        return new self([
+            'config'              => DeliveryOptionsConfig::fromCart($cart),
+            'cartShipmentOptions' => $cartShipmentOptions,
+            'settings'            => [
                 'hasDeliveryOptions' => $cart->shippingMethod->hasDeliveryOptions,
```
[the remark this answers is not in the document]
FreekVR: The double resolution is real but cheap, and the dedup doesn't depend on plugin caches: the capabilities repository stores per args-hash in PDK's own MemoryCacheStorage (request-scoped, bound in pdk-services.php, not overridden by Woo or Presta), and both passes run the same call sites on the same cart, so the cache keys are identical by construction. The second pass is one in-memory iteration over the carrier list — no repeated API calls within the request. Consolidating would either add mutable state to the service (a footgun for tests that change settings mid-run) or push carrier resolution up into the context model; neither seems worth it here.

src/App/DeliveryOptions/Service/DeliveryOptionsService.php
```diff
@@ -133,6 +145,72 @@ public function createAllCarrierSettings(PdkCart $cart): array
         return $settings;
     }
 
+    /**
+     * Calculate, per carrier, the shipment options this cart would be exported with, so the
+     * checkout can show and lock options that are already decided on the merchant side. Runs
+     * the same full calculation pipeline as an export (settings chain plus capabilities
+     * requires/excludes rules) on an order built from the cart, and converts the result to
+     * the capabilities key-space through the same converter the admin context uses.
+     *
```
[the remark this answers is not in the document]
FreekVR: Fixed in 3eb151a4. The docblock now describes the actual return value (calculated ShipmentOptions models) and points to the CheckoutContext boundary for the boolean conversion.

tests/Unit/App/Action/Backend/Order/ExportOrderActionTest.php
```diff
@@ -239,10 +239,12 @@ function exportWithSetting(bool $orderMode, CarrierSettingsFactory $carrierSetti
...
-    // Age check is enabled via export setting.
-    // Cascade behavior (age_check requires signature + only_recipient) is now driven
-    // by the capabilities API requires field, tested in CapabilitiesOptionCalculatorTest.
-    expect($options['age_check'])->toBe(1);
+    // Age check is enabled via the export setting. The capabilities requires/excludes
+    // relations (the permissive mock carries the real age verification rules) force
+    // signature and only recipient on with it, and receipt code off.
+    expect($options['age_check'])->toBe(1)
+        ->and($options['signature'])->toBe(1)
+        ->and($options['only_recipient'])->toBe(1);
```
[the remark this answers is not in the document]
FreekVR: Added in 3eb151a4.

src/App/DeliveryOptions/Service/DeliveryOptionsService.php
```diff
@@ -133,6 +145,75 @@ public function createAllCarrierSettings(PdkCart $cart): array
...
+    {
+        $order = new PdkOrder([
+            'lines'           => $cart->lines,
+            'shippingAddress' => $cart->shippingMethod->shippingAddress,
+            'deliveryOptions' => ['packageType' => $packageType],
+        ]);
+
+        $order->deliveryOptions->carrier = $carrier;
+
+        $calculatedOrder = $this->orderOptionsService->calculate($order);
```
NabDevs, line 230: Could this 'synthetic' order be initialized with empty notes?
I've looked at `calculate()` and it also runs the label description calculator. When [CUSTOMER_NOTE] is configured, it tries to load notes using this synthetic order, but the order has no ID. In Woo this throws an error and can potentially break the checkout i suppose.

NabDevs (changes requested):
1 thing to check, looks good otherwise.

src/App/DeliveryOptions/Service/DeliveryOptionsService.php
FreekVR: Spot on — thanks, this took the checkout down in my own test too. Fixed in c634d1ec: an order that only exists in memory now has no notes at all, so nothing asks the shop for them, and the cart's order is built with empty notes to say so. Tests use a note repository that rejects a missing identifier like Woo and Presta do, so it can't slip through unnoticed again. Notes turned out to be the only attribute that lazily loads from the shop, and only the full calculation touches it, so the blast radius was just this new checkout path.

src/App/DeliveryOptions/Service/DeliveryOptionsService.php
```diff
@@ -133,6 +145,78 @@ public function createAllCarrierSettings(PdkCart $cart): array
...
+    {
+        if (! $cart->shippingMethod->hasDeliveryOptions) {
+            return new ShipmentOptionsCollection();
+        }
+
+        [$packageType, $carriers] = $this->getValidCarrierOptions($cart);
+
+        $cartShipmentOptions = new ShipmentOptionsCollection();
+
+        foreach ($carriers as $carrier) {
```
GravendeelJochem, line 176: Zou je de per-carrier berekening in een try/catch willen zetten die logt en de carrier overslaat? Iets als:
`try {
    $cartShipmentOptions->put($identifier, $this->calculateCartShipmentOptions($carrier, $cart, $packageType));
} catch (Throwable $e) {
    Logger::error('Could not calculate cart shipment options', ['carrier' => $identifier, 'exception' => $e]);
}`

GravendeelJochem (changes requested):

NabDevs (dismissed):

src/App/DeliveryOptions/Service/DeliveryOptionsService.php
FreekVR: Good suggestion, added a generic try/catch and indicated why we don't consider this a fatal error.

GravendeelJochem (dismissed):

GravendeelJochem (approved):

FreekVR:
Closing this unmerged. INT-1596 is archived and the work continues in INT-1844, which takes a different approach: the capabilities API resolves requires and excludes itself when you send the options that are currently on, so the PDK stops doing that traversal at all.

Verified on production. A request with `options: {requiresAgeVerification: {}}` for PostNL, NL, PACKAGE returns `recipientOnlyDelivery` and `requiresSignature` as `isRequired: true` and drops `requiresReceiptCode` from the response entirely. Sending no carrier resolves every carrier independently in one response.

That makes `propagateConstraints`, `logConflict` and the key-collection helpers unnecessary, and with them the synthetic `PdkOrder` in `calculateCartShipmentOptions`, the null-id notes guard it needed, and the strict note repository double.

The branch stays, so nothing is lost. Two things to cherry-pick later: the 18+ carrier settings fix in `CarrierSettingsItemView.php:256-305`, where the lock reacts to any truthy tri-state instead of to ENABLED, which is an unrelated real bug and part of what INT-1596 reported; and the capability fixtures carrying real age-verification relations.

Plan: `docs/superpowers/plans/2026-09-16-shipment-options-refactor.md` in js-pdk.

PR #511, opened 2026-07-29
fix(insurance): read insurance limits from the flat capabilities format

FreekVR:
reads insurance limits from the flat capabilities format instead of the nested wrapper the API is removing.

The MyParcel API is dropping the old nested wrapper around insurance limits from the capabilities and contract definitions responses (Core API story SB-2770). The PDK read the limits through that wrapper, so insurance would have broken in the webshops the moment it disappeared.

Insurance limits now come from the flat min, max and default fields. Each limit stands on its own: no minimum means there is no floor, no maximum means the carrier sets no ceiling so a requested amount is kept as-is, and no default falls back to the minimum. For the carrier settings the "insure up to" dropdown needs a maximum to list amounts against, so without one it stays empty and logs a warning instead. Previously any missing limit was a fatal error, so this is also a crash fix for stored carrier data that predates the flat fields.

The API still sends both formats until after this ships. So that nothing starts depending on the old one again, the PDK now drops it as capabilities come in, which means stored carrier data and the admin only ever see the flat fields. That part is temporary and can go once the SDK is regenerated without the wrapper (INT-1696).

Note on ordering: this must be released before SB-2770 deploys, and the SDK must not be regenerated until the plugin releases are out. Plugin installs resolve `^11.0.0-beta.28`, so an SDK without the wrapper would hand an older PDK a model with no `getInsuredAmount()` and fatal before shops can update. The existing SDK floor already has the flat fields, so nothing needed bumping here.

Follow-ups in the same story: the admin still reads the nested shape in JS-PDK, and both plugins need a migration that refreshes stored carrier data.

Resolves INT-1697
Part of INT-1695


joerivanveen (approved):

src/App/Order/Calculator/General/InsuranceCalculator.php
```diff
@@ -144,15 +172,22 @@ private function fetchShipmentInsurance(Carrier $carrier): ?RefCapabilitiesRespo
     /**
      * Calculate insurance from carrier settings when the shipment option is set to INHERIT.
      *
+     * When the carrier advertises no maximum, the "insure up to" setting is the only ceiling
```
GravendeelJochem: Deze docblock en de code spreken elkaar tegen. Als de carrier geen max adverteert, geeft getAllowedInsuranceAmounts() een lege array, waardoor CarrierSettingsItemView::getExportInsuranceFields() (regels 606–612) vroegtijdig returnt en de "insure up to"-velden niet gerenderd worden de gebruiker kan die cap dus niet instellen. $maxInsuranceValue valt dan terug op de default 0, en min($validated, 0) maakt er 0 van.
Kunnen we kiezen: óf de velden ook renderen als er geen max is (dan werkt de cap echt), óf deze docblock-alinea weghalen?

GravendeelJochem (changes requested):

NabDevs (approved):

src/App/Order/Calculator/General/InsuranceCalculator.php
FreekVR: Clarified the docblocks and added new ones.

Without a max, its not possible to render the UI element for "insure up to" (a dropdown of tiers).

This doesnt actually occur in any API responses, but clarified it just defaults to the carrier min amount in that case to be safe.

GravendeelJochem (approved):

PR #513, opened 2026-07-30
fix(checkout): guard invalid carrier settings

NabDevs:
Prevents the checkout from crashing when carrier settings are missing or invalid. In that case, the delivery options service uses the default package type and returns no carriers. Includes regression coverage for missing and malformed settings.

Relates to: [Woo PR](https://github.com/myparcelnl/woocommerce/compare/feat/restore-carrier-settings-reactivation)

fixes INT-1730


src/App/DeliveryOptions/Service/DeliveryOptionsService.php
```diff
@@ -229,8 +230,14 @@ private function getBaseSettings(CarrierSettings $carrierSettings, PdkCart $cart
      */
     private function getValidCarrierOptions(PdkCart $cart): array
     {
+        $carrierSettings = Settings::get(CarrierSettings::ID);
+        $carrierSettings = is_array($carrierSettings) ? $carrierSettings : [];
+
+        if (empty($carrierSettings)) {
+            return [DeliveryOptions::DEFAULT_PACKAGE_TYPE_NAME, new CarrierCollection()];
+        }
```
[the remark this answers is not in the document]
NabDevs: Good catch. The null-coalescing accesses themselves are safe, but a malformed nested entry could still reach CarrierSettings::fromCarrier() later. I changed the guard to fail closed when any carrier entry is not an array instead of filtering only the local copy, because the original settings group is read again downstream.

tests/Unit/App/DeliveryOptions/Service/DeliveryOptionsServiceInvalidCarrierSettingsTest.php
```diff
@@ -0,0 +1,65 @@
...
+    ]));
+
+    expect($result['packageType'])->toBe(DeliveryOptions::DEFAULT_PACKAGE_TYPE_NAME)
+        ->and($result['carrierSettings'])->toBe([]);
+})->with([
+    'missing'       => [null],
+    'empty array'   => [[]],
+    'boolean value' => [false],
+    'string value'  => ['invalid'],
+]);
```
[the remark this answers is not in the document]
NabDevs: Added regression cases for both.

src/App/DeliveryOptions/Service/DeliveryOptionsService.php
```diff
@@ -229,8 +230,19 @@ private function getBaseSettings(CarrierSettings $carrierSettings, PdkCart $cart
      */
     private function getValidCarrierOptions(PdkCart $cart): array
     {
+        $carrierSettings = Settings::get(CarrierSettings::ID);
+
+        if (! is_array($carrierSettings) || empty($carrierSettings)) {
+            return [DeliveryOptions::DEFAULT_PACKAGE_TYPE_NAME, new CarrierCollection()];
+        }
+
+        foreach ($carrierSettings as $settings) {
```
FreekVR: check: if one carrier is missing / has malformed settings, no carriers are returned? shouldn't we filter out the carriers with invalid settings so that the rest can still pass the resolution logic below? Altenatively, the existing "$allCarriers" fallback at the end of the method seems more appropiate.

FreekVR (changes requested):

src/App/DeliveryOptions/Service/DeliveryOptionsService.php
NabDevs: Good point, fixed. I now only skip the malformed carrier setting, so the other valid carriers can still go through the normal resolution flow. I didn't use the $allCarriers fallback because that could also return disabled carriers.

joerivanveen (approved):

src/App/DeliveryOptions/Service/DeliveryOptionsService.php
```diff
@@ -229,8 +230,21 @@ private function getBaseSettings(CarrierSettings $carrierSettings, PdkCart $cart
...
+            return [DeliveryOptions::DEFAULT_PACKAGE_TYPE_NAME, new CarrierCollection()];
+        }
+
+        $carrierSettings = array_filter($carrierSettings, static function ($settings): bool {
+            return is_array($settings);
+        });
+
+        if (empty($carrierSettings)) {
+            return [DeliveryOptions::DEFAULT_PACKAGE_TYPE_NAME, new CarrierCollection()];
+        }
```
FreekVR, line 242: Nit (approving regardless): can be combined with nearly identical block at line 235, saves some code and makes it more readable.

FreekVR (approved):

src/App/DeliveryOptions/Service/DeliveryOptionsService.php
NabDevs: Ah fair, will implement this!

